PR #3, opened 2019-04-03
CEDS-1055: moved inventory linking api call to this service and refactored code so 

mattclark-zerogravit:
controller takes payload


app/uk/gov/hmrc/exports/movements/services/MovementsService.scala
```diff
@@ -0,0 +1,97 @@
...
+    movementsRepo
+      .save(movementSubmission)
+      .map(
+        res =>
+          if (res) {
+            Logger.debug("movement submission data saved to DB")
+            true
+          } else {
+            Logger.error("error  saving movement submission data to DB")
+            false
```
PatrykRudnicki:
```
if(res) Logger.debug("movement submission data saved to DB")
else Logger.error("error  saving movement submission data to DB")
res
```
Just proposal, I don't like if returning boolean :D

precheck.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-sbt clean coverage test it:test scalafmt::test test:scalafmt::test coverageReport
+sbt clean coverage test it:test coverageReport
```
PatrykRudnicki: Why you remove it? It's only checking and showing errors, we should have some style defined

PatrykRudnicki:
And please run formatter, few methods have wrong indentation

precheck.sh
mattclark-zerogravit: Every time I run precheck to see if my test are passing i get 20+ changes that I have to revert... yes to scalafmt not in precheck...

app/uk/gov/hmrc/exports/movements/services/MovementsService.scala
mattclark-zerogravit: done

PR #11, opened 2019-07-04
CEDS-1219 Fixed deprecated warnings and removed deprecated dependency

PatrykRudnicki:


test/integration/uk/gov/hmrc/exports/movements/services/MovementsServiceSpec.scala
```diff
@@ -74,35 +75,33 @@ class MovementsServiceSpec
         startInventoryLinkingService(ACCEPTED)
         withMovementSubmissionPersisted(true)
 
-        val result: Result = await(
+        val result: Future[Result] =
```
ed0906: Cant  we copy the await method over?

test/unit/uk/gov/hmrc/exports/movements/controllers/HeaderValidatorSpec.scala
```diff
@@ -19,7 +19,7 @@ package unit.uk.gov.hmrc.exports.movements.controllers
 import org.scalatest.mockito.MockitoSugar
 import uk.gov.hmrc.exports.movements.controllers.HeaderValidator
 import uk.gov.hmrc.exports.movements.models._
-import uk.gov.hmrc.play.test.UnitSpec
+import unit.uk.gov.hmrc.exports.movements.base.UnitSpec
```
ed0906: Potentially a Tech Debt task for the future but we can use source sets to differentiate between `test` and `integration` rather than having a package for each

test/integration/uk/gov/hmrc/exports/movements/services/MovementsServiceSpec.scala
PatrykRudnicki: `contentAsString` requires Future, so this is the reason that I changed to this approach. If you have some propose how to change `contentAsString` to something different let me know :)

test/unit/uk/gov/hmrc/exports/movements/controllers/HeaderValidatorSpec.scala
PatrykRudnicki: As I understand yesterday we want to have packages too

PatrykRudnicki: but I can be wrong of course 😂

PR #13, opened 2019-07-16
CEDS-1400 new persistence model

MaciejRewera:
This part contains changes for Notifications only.


app/uk/gov/hmrc/exports/movements/controllers/MovementSubmissionController.scala
```diff
@@ -21,26 +21,24 @@ import play.api.Logger
...
-class MovementsSubmissionController @Inject()(
+class MovementSubmissionController @Inject()(
   appConfig: AppConfig,
   authConnector: AuthConnector,
   headerValidator: HeaderValidator,
-  movementsService: MovementsService,
+  movementsService: MovementSubmissionService,
   cc: ControllerComponents
-) extends ExportController(authConnector, cc) {
+) extends Authenticator(authConnector, cc) {
```
ed0906: Minor but given inheritance should be read as "X is a Y" I think it would make more sense to name `Authenticator` `AuthenticatedController` (or similar)

e.g. 
`MovementSubmissionController` is an `AuthenticatedController` 
reads better than 
`MovementSubmissionController` is an `Authenticator` which doesn't really make sense

app/uk/gov/hmrc/exports/movements/models/notifications/MovementNotification.scala
```diff
@@ -0,0 +1,58 @@
...
+import play.api.libs.json.Json
+
+final case class MovementNotification(
+  dateTimeReceived: LocalDateTime = LocalDateTime.now(),
+  conversationId: String,
+  errors: Seq[NotificationError],
+  payload: String
+) {
+
+  override def equals(other: Any): Boolean = other match {
```
ed0906: One of the main benefits of Case classes is we don't need to implement equals & hashCode

How come we've done this?

app/uk/gov/hmrc/exports/movements/controllers/NotificationController.scala
```diff
@@ -0,0 +1,84 @@
...
+        } yield result
+
+        timer.stop()
+        savingNotificationResult.recover {
+          case exc: IllegalArgumentException =>
+            logger.error(s"There is a problem during parsing notification with exception: ${exc.getMessage}")
+            Accepted
+        }
+
+      case Left(errorResponse) => Future.successful(errorResponse.XmlResult)
```
ed0906: Do we not use the standard error handler in this service?

app/uk/gov/hmrc/exports/movements/services/NotificationService.scala
```diff
@@ -0,0 +1,42 @@
...
+@Singleton
+class NotificationService @Inject()(notificationRepository: NotificationRepository) {
+
+  private val logger = Logger(this.getClass)
+
+  def save(notification: MovementNotification): Future[Either[String, Unit]] =
+    notificationRepository
+      .insert(notification)
+      .map(_ => Right((): Unit))
+      .recover {
```
ed0906: Why not let it bubble up to the top? It saves us having to do `Left` and `Right` handling everywhere for exceptional circumstances

build.sbt
```diff
@@ -82,7 +82,7 @@ lazy val scoverageSettings: Seq[Setting[_]] = Seq(
     ,"metrics\\..*"
     ,".*(BuildInfo|Routes|Options).*"
   ).mkString(";"),
-  coverageMinimum := 95,
+  coverageMinimum := 90,
```
ed0906: :o

test/unit/uk/gov/hmrc/exports/movements/controllers/NotificationControllerSpec.scala
```diff
@@ -0,0 +1,228 @@
...
+        val result = routePostSaveNotification()
+
+        status(result) must be(ACCEPTED)
+      }
+
+      "call MovementNotificationFactory and NotificationService afterwards" in new HappyPathSaveControlResponseTest {
+
+        routePostSaveNotification().futureValue
+
+        val inOrder: InOrder = Mockito.inOrder(movementNotificationFactoryMock, notificationServiceMock)
```
ed0906: Whats the value of using InOrder here? Does it matter what the order is?

test/unit/uk/gov/hmrc/exports/movements/models/notifications/MovementNotificationSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+import org.scalatest.{MustMatchers, WordSpec}
+import uk.gov.hmrc.exports.movements.models.notifications.{MovementNotification, NotificationError}
+
+class MovementNotificationSpec extends WordSpec with MustMatchers {
+
+  import MovementNotificationSpec._
+
+  "MovementNotification on equals" should {
+
+    "be Reflexive" in {
```
ed0906: Depending on the outcome of the other comments we might not need this

app/uk/gov/hmrc/exports/movements/models/notifications/MovementNotification.scala
MaciejRewera: This is a long story...
We need to have some chronology in our Notifications, but the payload does not contain any timestamps. Therefore, we have to add it ourselves.
But on the other hand if we receive the same Notification twice, we will create different receivedTime. So we should not rely upon this field when comparing Notifications.

app/uk/gov/hmrc/exports/movements/models/notifications/MovementNotification.scala
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+package uk.gov.hmrc.exports.movements.models.notifications
+
+import java.time.LocalDateTime
+
+import play.api.libs.json.Json
+
+final case class MovementNotification(
+  dateTimeReceived: LocalDateTime = LocalDateTime.now(),
```
ed0906: I presume we don't need to care about timezones? Generally `Instant` is the recommended way to store dates (UTC time)

MaciejRewera: I am not sure about this, as schemas can handle different time zones. Exports are international stuff, so it might be the case.

test/unit/uk/gov/hmrc/exports/movements/controllers/NotificationControllerSpec.scala
MaciejRewera: Yes, it does.

app/uk/gov/hmrc/exports/movements/services/NotificationService.scala
MaciejRewera: In some cases when we have exceptions thrown, we don't want the client service to know about the problems, so we just log the problem but still return Accepted status code.
This is also how we do it on exports BE, so I wanted to be concise.

app/uk/gov/hmrc/exports/movements/controllers/NotificationController.scala
MaciejRewera: We don't have them on backends.

app/uk/gov/hmrc/exports/movements/controllers/MovementSubmissionController.scala
MaciejRewera: You're right. I'll change that. The initial idea was not to extend this class, but to compose it into Controller. But it is out of this ticket's scope.

ed0906: Sure :)

app/uk/gov/hmrc/exports/movements/controllers/NotificationController.scala
mattclark-zerogravit: To give some context, customs-notification can't  / wont handle any errors and sending back anything other than accepted could block subsequent notifications os this is fine

app/uk/gov/hmrc/exports/movements/models/notifications/MovementNotification.scala
ed0906: `LocalDateTime` is just a Date with a Time e.g. `2019-07-17T11:05:00` so we need to be careful. `LocalDateTime.now()` will be the date + time in the timezone of the server it is run on, is this okay?

app/uk/gov/hmrc/exports/movements/controllers/NotificationController.scala
```diff
@@ -0,0 +1,84 @@
...
+    val timer = metrics.startTimer(movementMetric)
+    headerValidator.validateAndExtractMovementNotificationHeaders(request.headers.toSimpleMap) match {
+      case Right(extractedHeaders) =>
+        val savingNotificationResult = for {
+          notificationToSave <- buildNotificationFromResponse(extractedHeaders.conversationId.value, request.body)
+          serviceResponse <- notificationService.save(notificationToSave)
+          result = handleServiceResponse(serviceResponse)
+        } yield result
+
+        timer.stop()
```
PauloPortugal: Will the timer continue running if we have a `Left`

app/uk/gov/hmrc/exports/movements/controllers/NotificationController.scala
ed0906: In that case do we want to be sending `errorResponse` back?

app/uk/gov/hmrc/exports/movements/models/notifications/MovementNotification.scala
PauloPortugal: So from what I understand here, we do not want to include the `dateTimeReceived` as part of our equality contract.
This is to distinguish different messages/notifications sent by ILE, correct?

Did you use IntelliJ's generated code approach?

app/uk/gov/hmrc/exports/movements/models/notifications/MovementNotificationFactory.scala
```diff
@@ -0,0 +1,51 @@
...
+
+@Singleton
+class MovementNotificationFactory {
+
+  private val logger = Logger(this.getClass)
+
+  private val inventoryLinkingControlResponseLabel = "inventoryLinkingControlResponse"
+  private val inventoryLinkingMovementTotalsResponse = "inventoryLinkingMovementTotalsResponse"
+
+  def buildMovementNotification(conversationId: String, xml: NodeSeq): MovementNotification =
```
PauloPortugal: :+1:

build.sbt
PauloPortugal: Ideally, we should not drop the code coverage. Are there areas of the code that we should be covering these?
Probably if we follow Ed's recommendations on allowing the exceptions to bubble up and we remove the `Left` and `Right` we may increase some code coverage and avoid more complex logic.

test/unit/uk/gov/hmrc/exports/movements/controllers/NotificationControllerSpec.scala
```diff
@@ -0,0 +1,228 @@
...
+      }
+
+      "call MovementNotificationFactory once, " +
+        "passing conversationId from headers and request body" in new HappyPathSaveControlResponseTest {
+
+        routePostSaveNotification().futureValue
+
+        val conversationIdCaptor: ArgumentCaptor[String] = ArgumentCaptor.forClass(classOf[String])
+        val requestBodyCaptor: ArgumentCaptor[Elem] = ArgumentCaptor.forClass(classOf[Elem])
+        verify(movementNotificationFactoryMock, times(1))
```
PauloPortugal: `verify(movementNotificationFactoryMock, times(1))` is the same as `verify(movementNotificationFactoryMock)`
Not a big thing at all, just a comment. :wink:

test/utils/NotificationTestData.scala
```diff
@@ -18,117 +18,113 @@ package utils
...
-  val uri = "/customs-declare-exports/notify"
   val movementUri = "/customs-declare-exports/notifyMovement"
-  val submissionNotificationUri = "/customs-declare-exports/submission-notifications/1234"
 
-  val getNotificationUri = "/customs-declare-exports/notifications"
-  val postNotificationUri = "/customs-declare-exports/notify"
-  val validXML: Elem = <MetaData xmlns="urn:wco:datamodel:WCO:DocumentMetaData-DMS:2">
-    <wstxns1:Response xmlns:wstxns1="urn:wco:datamodel:WCO:RES-DMS:2"></wstxns1:Response>
-  </MetaData>
+  val exampleRejectInventoryLinkingControlResponseNotification: MovementNotification = MovementNotification(
```
PauloPortugal: :+1:

app/uk/gov/hmrc/exports/movements/models/notifications/MovementNotification.scala
PauloPortugal: Ahhhhh!!!! Really good comment. Yes we should be using `UTC` on our DBs, but displaying the timestamps in their local time.
I think we should have that conversation separately, because it would be good to align FE Exports and FE Movements with the same strategy

app/uk/gov/hmrc/exports/movements/models/notifications/MovementNotification.scala
```diff
@@ -0,0 +1,58 @@
...
+    result
+  }
+
+}
+
+object MovementNotification {
+  implicit val format = Json.format[MovementNotification]
+
+  def empty = MovementNotification(
+    conversationId = "",
```
mattclark-zerogravit: I'd suggest we store the type of the notification somehow here... as different actions to ILE can trigger multiple notifications of different types

build.sbt
mattclark-zerogravit: Yeah this is not good..should not be reducing coverage..

app/uk/gov/hmrc/exports/movements/connectors/CustomsInventoryLinkingExportsConnector.scala
```diff
@@ -66,7 +67,7 @@ class CustomsInventoryLinkingExportsConnector @Inject()(appConfig: AppConfig, ht
       }
   }
 
-  // TODO: changed from hardcoded values to the ones from object, as it was before set to "Identfier"
+  // TODO: changed from hardcoded values to the ones from object, as it was before set to "Identifier"
```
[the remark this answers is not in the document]
MaciejRewera: It was there before us... I'll remove it.

app/uk/gov/hmrc/exports/movements/controllers/NotificationController.scala
MaciejRewera: Good catch! Will change that.

build.sbt
MaciejRewera: After speaking to @mattclark-zerogravit I'll leave the coverage as it is, but created tech-debt ticket to remember about it.

app/uk/gov/hmrc/exports/movements/controllers/NotificationController.scala
MaciejRewera: Changed to return Accepted status and log the error.

app/uk/gov/hmrc/exports/movements/models/notifications/MovementNotification.scala
MaciejRewera: I'm going to get rid of custom equals and hashCode. The equality of Notifications shall be handled in separate tickets somewhere in tthe future.

PR #15, opened 2019-07-23
CEDS-1417 Endpoint to retrieve notifications

PatrykRudnicki:


test/unit/uk/gov/hmrc/exports/movements/controllers/NotificationControllerSpec.scala
```diff
@@ -210,7 +210,6 @@ class NotificationControllerSpec
         status(result) must be(ACCEPTED)
       }
     }
-
```
PatrykRudnicki: I have a huge problem with migrating those tests ^^ to not use application, just to be unit spec, problem with xml validation or something like this. I tested piece of code that I added in another file with unit approach and will try to migrate this in the nearest future.

PR #19, opened 2019-07-29
Removed wrong TODO

PatrykRudnicki:


app/uk/gov/hmrc/exports/movements/models/notifications/NotificationFactory.scala
```diff
@@ -20,7 +20,6 @@ import javax.inject.Singleton
 
 import scala.xml.NodeSeq
 
-//TODO This should be an object. There is not point to need to inject this or have it as class. Changing to object will break a lot of code, so be careful
 @Singleton
 class NotificationFactory {
```
ed0906: I think if you just stuck the `@Inject()()` it would be obvious enough it might be used as a dependency somewhere

PatrykRudnicki: Added, thanks!

PR #26, opened 2019-08-13
CEDS-1436 ILE response validation

MaciejRewera:


app/uk/gov/hmrc/exports/movements/config/AppConfig.scala
```diff
@@ -25,13 +25,20 @@ class AppConfig @Inject()(runModeConfiguration: Configuration, servicesConfig: S
 
   lazy val authUrl: String = servicesConfig.baseUrl("auth")
 
-  lazy val customsInventoryLinkingExports = servicesConfig.baseUrl("customs-inventory-linking-exports")
-  lazy val sendArrival = servicesConfig.getConfString(
+  lazy val customsInventoryLinkingExportsUrl = servicesConfig.baseUrl("customs-inventory-linking-exports")
+
+  lazy val sendArrivalUrlSuffix = servicesConfig.getConfString(
```
ed0906: “path” is probably more descriptive rather than “suffix”

app/uk/gov/hmrc/exports/movements/config/AppConfig.scala
```diff
@@ -25,13 +25,20 @@ class AppConfig @Inject()(runModeConfiguration: Configuration, servicesConfig: S
 
   lazy val authUrl: String = servicesConfig.baseUrl("auth")
 
-  lazy val customsInventoryLinkingExports = servicesConfig.baseUrl("customs-inventory-linking-exports")
-  lazy val sendArrival = servicesConfig.getConfString(
+  lazy val customsInventoryLinkingExportsUrl = servicesConfig.baseUrl("customs-inventory-linking-exports")
```
ed0906: Maybe “root” instead of “url”? Minor though

app/uk/gov/hmrc/exports/movements/models/notifications/NotificationFactory.scala
```diff
@@ -35,4 +43,13 @@ class NotificationFactory @Inject()(responseParserFactory: ResponseParserFactory
     )
   }
 
+  private def checkResponseCompliance(conversationId: String, xml: NodeSeq): Unit =
+    responseValidator.validate(xml).recover {
```
SteveSugden: Based on similar code I changed yesterday, does this not "swallow" the exception?  Is that what we want?

test/unit/uk/gov/hmrc/exports/movements/base/UnitTestMockBuilder.scala
```diff
@@ -137,4 +134,12 @@ object UnitTestMockBuilder extends MockitoSugar {
 
     responseParserMock
   }
+
+  def buildResponseValidatorMock: ResponseValidator = {
+    val responseValidatorMock = mock[ResponseValidator]
+
+    when(responseValidatorMock.validate(any[NodeSeq])).thenReturn(Try((): Unit))
```
PatrykRudnicki: I think mock instead of `Try` should use `Success` or `Failure`. But I'm not sure that there is some value from this, just probably more readable for understanding, because from the beginning you'll know that you're mocking success case or failure.

app/uk/gov/hmrc/exports/movements/models/notifications/NotificationFactory.scala
MaciejRewera: Yes. We only want to log the problem. The whole process should continue.

test/unit/uk/gov/hmrc/exports/movements/base/UnitTestMockBuilder.scala
MaciejRewera: 👍

MaciejRewera:
Test this please

PR #27, opened 2019-08-21
CEDS-1509 Add endpoint for getting single Submission

MaciejRewera:


MaciejRewera:
Test this please.

conf/app.routes
```diff
@@ -8,6 +8,7 @@ POST        /movements/arrival                              uk.gov.hmrc.exports.
 POST        /movements/departure                            uk.gov.hmrc.exports.movements.controllers.SubmissionController.submitDeparture()
 
 GET         /movements                                      uk.gov.hmrc.exports.movements.controllers.SubmissionController.getAllSubmissions()
+GET         /movements/:conversationId                      uk.gov.hmrc.exports.movements.controllers.SubmissionController.getSubmission(conversationId: String)
```
ed0906: Is there a way we could have our own ID instead of using `conversationId`?

Using `conversationId` in exports has caused quite  a lot of issues and we agreed in the techtro a while ago to try and replace it with our own ID

app/uk/gov/hmrc/exports/movements/controllers/SubmissionController.scala
```diff
@@ -88,8 +88,15 @@ class SubmissionController @Inject()(
         case Left(errorMsg) => ErrorResponse.errorInternalServerError(errorMsg).XmlResult
       }
 
-  def getAllSubmissions: Action[AnyContent] =
-    authorisedAction(parse.default) { implicit authorizedRequest =>
-      submissionService.getSubmissionsByEori(authorizedRequest.eori.value).map(movements => Ok(Json.toJson(movements)))
-    }
+  def getAllSubmissions: Action[AnyContent] = authorisedAction(parse.default) { implicit authorizedRequest =>
+    submissionService.getSubmissionsByEori(authorizedRequest.eori.value)
+      .map(movements =>
```
PatrykRudnicki: Just style, Ok(Json.toJson(movements)))` should be in the same line, like below `getSubmission` :)

app/uk/gov/hmrc/exports/movements/models/notifications/NotificationContract.scala
```diff
@@ -20,18 +20,18 @@ import java.time.Instant
 
 import play.api.libs.json.Json
 
-final case class NotificationPresentation(
+final case class NotificationContract(
```
PatrykRudnicki: It's not good name for NotificationModel that is shared between services.

IMO just `Notification`, `NotificationModel`, `NotificationFrontendModel`, `NotificationViewModel`, anything, contract is not something that I'll understand in this case

test/unit/uk/gov/hmrc/exports/movements/controllers/SubmissionControllerSpec.scala
```diff
@@ -55,10 +55,11 @@ class SubmissionControllerSpec
   private val arrivalUri = "/movements/arrival"
   private val departureUri = "/movements/departure"
   private val getAllSubmissionsUri = "/movements"
+  private def getSubmissionUri(conversationId: String) = s"/movements/$conversationId"
 
   private val submissionServiceMock = buildSubmissionServiceMock
-  private val metrics: MovementsMetrics = app.injector.instanceOf[MovementsMetrics]
-  private val headerValidator: HeaderValidator = app.injector.instanceOf[HeaderValidator]
+  implicit private val metrics: MovementsMetrics = app.injector.instanceOf[MovementsMetrics]
```
PatrykRudnicki: Hmm, why you need to have implicit metrics and implicit headerValidator?

app/uk/gov/hmrc/exports/movements/controllers/SubmissionController.scala
MaciejRewera: scalafmt sometimes does not make changes in such situations...

PatrykRudnicki: it's not solving correct cases, this case is correct :)

test/unit/uk/gov/hmrc/exports/movements/controllers/SubmissionControllerSpec.scala
```diff
@@ -239,8 +240,7 @@ class SubmissionControllerSpec
 
     "return Ok status" in {
       withAuthorizedUser()
-      when(submissionServiceMock.getSubmissionsByEori(any[String]))
-        .thenReturn(Future.successful(Seq(exampleSubmission())))
+      when(submissionServiceMock.getSubmissionsByEori(any[String])).thenReturn(Future.successful(Seq.empty))
```
PatrykRudnicki:
```
private val metrics: MovementsMetrics = app.injector.instanceOf[MovementsMetrics]
private val headerValidator: HeaderValidator = app.injector.instanceOf[HeaderValidator]
```
This is not used, but it's not your code. Feel free to remove it.

PR #30, opened 2019-08-29
CEDS-1552 Changed logger approach, enable scalafmt fail and reformat codebase

PatrykRudnicki:


build.sbt
```diff
@@ -36,8 +36,7 @@ lazy val microservice = Project(appName, file("."))
     dependencyOverrides ++= AppDependencies.jettyOverrides,
     evictionWarningOptions in update := EvictionWarningOptions.default.withWarnScalaVersionEviction(false),
     majorVersion := 0,
-    scalaVersion := "2.12.8",
-    scalafmtFailTest in ThisBuild := false
```
ed0906: What's this do?

PatrykRudnicki: Okay, it's missing thing, we removed this from frontend and exports, it'll provide failing build when code will be not formatted.

PR #32, opened 2019-09-10
CEDS-1602 Logging review

bambuchaAdm:
Plus service return signature refactor


app/uk/gov/hmrc/exports/movements/connectors/CustomsInventoryLinkingExportsConnector.scala
```diff
@@ -64,8 +64,7 @@ class CustomsInventoryLinkingExportsConnector @Inject()(appConfig: AppConfig, ht
       )(responseReader, hc, ec)
       .recover {
         case error: Throwable =>
-          logger.error(s"Error to check development environment ${error.toString}")
-          logger.error(s"Error to check development environment (GET MESSAGE) ${error.getMessage}")
+          logger.warn(s"Error to check development environment. $error")
```
ed0906: I think we should improve the message as it doesnt really make sense, maybe something like:
"Bad Response from Customs Inventory Linking [$error]"

app/uk/gov/hmrc/exports/movements/controllers/ConsolidationController.scala
```diff
@@ -60,7 +60,7 @@ class ConsolidationController @Inject()(
         parse.tolerantXml(rq).map {
           case Right(xml) => Right(AnyContentAsXml(xml))
           case _ =>
-            logger.error("Invalid xml payload")
+            logger.warn("Invalid xml payload")
```
ed0906: If we saw this in the logs I'm not sure how we would act on it, I think we need more context, e.g. why it failed

app/uk/gov/hmrc/exports/movements/controllers/ConsolidationController.scala
```diff
@@ -74,16 +74,19 @@ class ConsolidationController @Inject()(
           SubmissionRequestContext(eori = request.eori.value, actionType = actionType, requestXml = requestXml)
         forwardMovementConsolidationRequest(context)
       case None =>
-        logger.error("Body is not xml")
+        logger.warn("Body is not xml")
```
ed0906: I think we need more context in the error message again so we can tell what the error is better when looking at the logs e.g 
"Bad Consolidation Request: Invalid XML"

app/uk/gov/hmrc/exports/movements/controllers/NotificationController.scala
```diff
@@ -70,17 +71,14 @@ class NotificationController @Inject()(
       Some(notificationFactory.buildMovementNotification(conversationId, responseXml))
     } catch {
       case exc: IllegalArgumentException =>
-        logger.error(s"There is a problem during parsing notification with exception: ${exc.getMessage}")
+        logger.warn(s"There is a problem during parsing notification with exception: ${exc.getMessage}")
```
ed0906: Sounds to me like an `error`  rather than a `warn`. If we start receiving invalid notifications we probably want to know about it

app/uk/gov/hmrc/exports/movements/controllers/actions/AuthenticatedController.scala
```diff
@@ -56,7 +55,9 @@ class AuthenticatedController @Inject()(override val authConnector: AuthConnecto
       hasEnrolment(enrolments) match {
         case Some(eori) =>
           Future.successful(Right(AuthorizedSubmissionRequest(Eori(eori.value), request)))
-        case _ => Future.successful(Left(ErrorResponse.ErrorUnauthorized))
+        case _ =>
+          logger.warn("User is does not have EORI number in enrollments")
```
ed0906: "User attempted to access Service with the expected role but without an EORI"

app/uk/gov/hmrc/exports/movements/controllers/NotificationController.scala
bambuchaAdm: According to logging guideline errors is mostly things which does not allow application work.
I think we should monitor number of those warnings, incidents. Sounds more like metric.

app/uk/gov/hmrc/exports/movements/controllers/util/HeaderValidator.scala
```diff
@@ -31,7 +31,7 @@ class HeaderValidator {
     headers.get(headerName) match {
       case Some(header) if !header.isEmpty => Some(header)
       case _ =>
-        logger.error(s"Error Extracting $headerName")
+        logger.warn(s"Error Extracting $headerName")
```
ed0906: "Expected Header [$headerName}] but it was missing/empty"

app/uk/gov/hmrc/exports/movements/models/notifications/NotificationFactory.scala
```diff
@@ -46,10 +47,12 @@ class NotificationFactory @Inject()(
   private def checkResponseCompliance(conversationId: String, xml: NodeSeq): Unit =
     responseValidator.validate(xml).recover {
       case exc: SAXParseException =>
+        MDC.put("conversationId", conversationId)
         logger
-          .error(
+          .warn(
             s"Received Notification for Conversation ID: [$conversationId] does not match the schema: ${exc.getMessage}"
           )
+        MDC.remove("conversationId")
```
ed0906: Why use `MDC`?

app/uk/gov/hmrc/exports/movements/services/NotificationService.scala
```diff
@@ -29,17 +29,10 @@ class NotificationService @Inject()(
   submissionRepository: SubmissionRepository
 )(implicit executionContext: ExecutionContext) {
 
-  private val logger = Logger(this.getClass)
-
-  def save(notification: Notification): Future[Either[String, Unit]] =
+  def save(notification: Notification): Future[Unit] =
```
ed0906: 👍

MaciejRewera: :+1:

app/uk/gov/hmrc/exports/movements/services/SubmissionService.scala
```diff
@@ -64,3 +63,12 @@ class SubmissionService @Inject()(
     submissionRepository.findByConversationId(conversationId)
 
 }
+
+class CustomsInventoryLinkingUpstreamException(status: Int, coversationId: Option[String], message: String)
```
MaciejRewera: I would prefer this exception to be in a dedicated package, in separate file.

app/uk/gov/hmrc/exports/movements/services/SubmissionService.scala
```diff
@@ -64,3 +63,12 @@ class SubmissionService @Inject()(
...
+    extends Exception(message) {
+  override def getMessage: String = {
+    val formattedConversation = coversationId.map(id => s"'$id'").getOrElse("Not preset")
```
MaciejRewera: Please add 'Id' to the name.

test/unit/uk/gov/hmrc/exports/movements/base/AuthTestSupport.scala
```diff
@@ -29,13 +30,18 @@ import utils.testdata.CommonTestData.validEori
 
 import scala.concurrent.Future
 
-trait AuthTestSupport extends MockitoSugar {
+trait AuthTestSupport extends MockitoSugar with BeforeAndAfterEach { self: Suite =>
```
MaciejRewera: Could you briefly explain usage of `self: Suite =>` ?

PatrykRudnicki: From Suite documentation:
```
A suite of tests. A <code>Suite</code> instance encapsulates a conceptual suite (i.e., a collection) of tests.

This trait provides an interface composed of "lifecycle methods" that allow suites of tests to be run.
Its implementation enables a default way of writing and executing tests.  Subtraits and subclasses can override <code>Suite</code>'s lifecycle methods to enable other ways of writing and executing tests.
```

I think this is the mail reason ^^ "encapsulation". But it's to wait for @bambuchaAdm to confirm

bambuchaAdm: BeforeAndAfterEach is self annotated to `Suite` when i extend it i have to have same self type annotation as it

PR #33, opened 2019-09-11
Clear metric registry in integration tests

bambuchaAdm:


test/integration/uk/gov/hmrc/exports/movements/repositories/SubmissionRepositorySpec.scala
```diff
@@ -35,7 +35,10 @@ class SubmissionRepositorySpec
     extends WordSpec with GuiceOneAppPerSuite with BeforeAndAfterEach with ScalaFutures with MustMatchers
     with IntegrationPatience {
 
-  override lazy val app: Application = GuiceApplicationBuilder().build()
+  override def fakeApplication: Application = {
```
PatrykRudnicki: I'm not sure that there is some value to change name to `fakeApplication`.

If you want to use `fake` in name, my propose is to use `fakeApp`.

`fakeApplication` is too long in my opinion, places where you're using this it's in route method where you have, sometimes some json etc, this change will break everything to few lines after formatting. I prefer to stay with `app`, but `fakeApp` is acceptable for me.

PR #36, opened 2019-09-19
CEDS-1639 Enhance NotificationFrontendModel

MaciejRewera:
More data is required to be presented to the user. ROE and SOE codes
need to be sent together with associated UCR.

### DIFF
--- a/app/uk/gov/hmrc/exports/movements/models/notifications/NotificationFrontendModel.scala
+++ b/app/uk/gov/hmrc/exports/movements/models/notifications/NotificationFrontendModel.scala
@@ -24,9 +24,7 @@ final case class NotificationFrontendModel(
   timestampReceived: Instant = Instant.now(),
   conversationId: String,
   responseType: String,
-  ucrBlocks: Seq[UcrBlock],
-  masterRoe: Option[String],
-  masterSoe: Option[String],
+  entries: Seq[Entry],
   crcCode: Option[String],
   actionCode: Option[String],
   errorCodes: Seq[String]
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/exports/movements/models/notifications/NotificationFrontendModel.scala
```diff
@@ -35,16 +33,26 @@ final case class NotificationFrontendModel(
 object NotificationFrontendModel {
   implicit val format = Json.format[NotificationFrontendModel]
 
-  def apply(notification: Notification): NotificationFrontendModel =
+  def apply(notification: Notification): NotificationFrontendModel = {
+    val mucrEntry = for {
```
SteveSugden: Does this have to done using a for-comprehension?  Can you not just construct the Entry from the data?

app/uk/gov/hmrc/exports/movements/models/notifications/NotificationFrontendModel.scala
```diff
@@ -35,16 +33,26 @@ final case class NotificationFrontendModel(
...
+      ucr <- notification.data.masterUcr
```
bambuchaAdm: maybe having delegating method here would be more intuitive - this code does not have to know that notification has internal data structure

app/uk/gov/hmrc/exports/movements/models/notifications/NotificationFrontendModel.scala
PatrykRudnicki: What about ducr?

MaciejRewera: DUCRs are stored in entries already. The extra codee here is to align MUCR handling for us, instead of having extra fields as it is in the payload.

MaciejRewera: Actually I could. Thank you for the idea!

app/uk/gov/hmrc/exports/movements/models/notifications/NotificationFrontendModel.scala
MaciejRewera: Added methods to Notification class.

test/unit/uk/gov/hmrc/exports/movements/models/notifications/NotificationFrontendModelSpec.scala
```diff
@@ -0,0 +1,105 @@
...
+
+class NotificationFrontendModelSpec extends WordSpec with MustMatchers {
+
+  "NotificationFrontendModel on apply" when {
+
+    "provided with Notification" which {
+
+      "contains no MUCR or DUCR" should {
+        "return NotificationFrontendModel without entries" in {
+          val notification = notification_1
```
PatrykRudnicki: I'm not sure that there is value of new val here

PR #38, opened 2019-09-20
CEDS-1667 Support for CHIEF errors

PatrykRudnicki:


test/unit/uk/gov/hmrc/exports/movements/models/notifications/parsers/ErrorSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+class ErrorSpec extends UnitSpec {
+
+  "Error apply method" should {
+
+    "return an Error when the structure is correct" in {
+
+      val inputList = List("code", "description")
+      val expectedError = Error("code", "description")
+
+      Error.apply(inputList) shouldBe expectedError
```
ed0906: `Error(inputList)`?

PatrykRudnicki: Fixed

app/uk/gov/hmrc/exports/movements/models/notifications/parsers/ErrorParser.scala
```diff
@@ -0,0 +1,82 @@
...
+package uk.gov.hmrc.exports.movements.models.notifications.parsers
+
+import java.util.regex.Pattern
+
+import com.github.tototoshi.csv._
+import play.api.Logger
+
+import scala.io.Source
+
+case class Error(code: String, description: String)
```
MaciejRewera: We should not use descriptionthis way. It is better to use message keys, as we will eventually need to implement Welsh.

app/uk/gov/hmrc/exports/movements/models/notifications/parsers/ErrorParser.scala
```diff
@@ -0,0 +1,82 @@
...
+  private val chiefErrors: List[Error] = {
+    val source = Source.fromURL(getClass.getClassLoader.getResource("chief_errors.csv"), "UTF-8")
+
+    readErrorsFromFile(source)
+  }
+
+  private def retrieveChiefErrorCode(errorMessage: String): Option[String] =
+    errorMessage.split(" ").filter(chiefErrorPattern.matcher(_).matches).headOption
+
+  def validateErrors(errors: Seq[String]): Seq[String] = {
```
MaciejRewera: As per 'Clean code' guidance, it is better to have public methods first and private methods after.

app/uk/gov/hmrc/exports/movements/models/notifications/parsers/ErrorParser.scala
PatrykRudnicki: So we're reading descriptions from the file in that case it's doesn't make any sense to rewrite it to messages.en. I'll add sticker about it to the board, and we can discuss it

PR #39, opened 2019-09-23
CEDS-1667 Proper naming for Error Validator

PatrykRudnicki:


app/uk/gov/hmrc/exports/movements/models/notifications/parsers/ErrorValidator.scala
```diff
@@ -37,12 +37,24 @@ object Error {
   }
 }
 
-object ErrorParser {
+object ErrorValidator {
+
+  def validateErrors(errors: Seq[String]): Seq[String] = {
+
+    val correctIleErrors = errors.filter(ileErrors.map(_.code).contains(_))
```
bambuchaAdm: `ileErrors` is quite stable  - are we able to provide somewhere (ex. in companion object) `Map[Code, Error]` - would simplify much this lookup ?

app/uk/gov/hmrc/exports/movements/models/notifications/parsers/ErrorValidator.scala
```diff
@@ -37,12 +37,24 @@ object Error {
...
+
+    val retrievedChiefErrors =
+      errors.diff(correctIleErrors).map(retrieveChiefErrorCode).filter(_.isDefined).map(_.get)
```
bambuchaAdm: `filter(_.isDefined).map(_.get)` => `flatten`

PatrykRudnicki: Good point, fixed

test/unit/uk/gov/hmrc/exports/movements/models/notifications/parsers/ErrorValidatorSpec.scala
```diff
@@ -28,7 +28,7 @@ class ErrorParserSpec extends UnitSpec {
       val errors = Seq("02", "09", "15", "20", "30", "41", "E607", "E1236549", "E10419", "E3464")
       val correctErrors = Seq("02", "15", "20", "30", "E607", "E10419", "E3464")
 
-      ErrorParser.validateErrors(errors) shouldBe correctErrors
+      ErrorValidator.validateErrors(errors) shouldBe correctErrors
```
ed0906: Do we want to be able to mock out the `ErrorValidator` where it is used? If so I would  suggest we make it a `class` rather than `object`

app/uk/gov/hmrc/exports/movements/models/notifications/parsers/ErrorValidator.scala
ed0906: I agree, `ileErrors` sounds a better name to me

app/uk/gov/hmrc/exports/movements/models/notifications/parsers/ErrorValidator.scala
```diff
@@ -37,7 +37,25 @@ object Error {
   }
 }
 
-object ErrorParser {
+object ErrorValidator {
+
+  private val logger = Logger(this.getClass)
+
+  def validateErrors(errors: Seq[String]): Seq[String] = {
```
ed0906: I think we could name this differently as it is more of a  `filter` than a `validate` (if you agree)

I don't know the context it is used in, but if we have a `Seq[String]` it might be nice to do

```
val errors = Seq(...)
val chiefErrors = errors.filter(ErrorValidator.chiefOnly)
val ileErrors =  errors.filter(ErrorValidator.ileOnly)
```

app/uk/gov/hmrc/exports/movements/models/notifications/parsers/ErrorValidator.scala
PatrykRudnicki: Ah, but we have `ileErrors` already defined. It contains all correct errors from csv.

app/uk/gov/hmrc/exports/movements/models/notifications/parsers/ErrorValidator.scala
PatrykRudnicki: Added in separate commit, please rereview.

PR #40, opened 2019-10-03
CEDS-1767 New endpoint to handle all consolidations

PatrykRudnicki:


app/uk/gov/hmrc/exports/movements/controllers/ConsolidationController.scala
```diff
@@ -81,4 +82,14 @@ class ConsolidationController @Inject()(
           ErrorResponse.errorInternalServerError(e.getMessage).XmlResult
       }
 
+  def submitConsolidation(): Action[ConsolidationRequest] = authorisedAction(parse.json[ConsolidationRequest]) {
+    implicit request =>
+      consolidationService
+        .submitConsolidationRequest(request.eori.value, request.body)
+        .map(_ => Accepted("Consolidation request submitted successfully"))
+        .recover {
```
ed0906: I don't think we need this, it should be handled by the error handler
We definitely shouldn't be sending XML payloads out the door either way

app/uk/gov/hmrc/exports/movements/models/consolidation/ConsolidationRequest.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  implicit val format = Json.format[ConsolidationRequest]
+
+  def apply(consolidationType: ConsolidationType, mucr: Option[String], ducr: Option[String]): ConsolidationRequest =
+    consolidationType match {
+      case AssociateDucr if mucr.isDefined && ducr.isDefined => new ConsolidationRequest(AssociateDucr, mucr, ducr)
+      case DisassociateDucr if ducr.isDefined                => new ConsolidationRequest(DisassociateDucr, None, ducr)
+      case ShutMucr if mucr.isDefined                        => new ConsolidationRequest(ShutMucr, mucr, None)
+      case _: ConsolidationType                              => throw new IllegalArgumentException("Missing mandatory information")
+    }
```
ed0906: I think you could achieve the same thing using inheritance, that way Play would do the validation for you

ed0906:
```
trait ConsolidationRequest {
    `type`: ConsolidationType`
}

case class AssociateDucrRequest(mucr: String, ducr: String) extends ConsolitationRequest {
    override val `type` = AssociateDucr
}
```

app/uk/gov/hmrc/exports/movements/controllers/ConsolidationController.scala
PatrykRudnicki: Good point, removed

app/uk/gov/hmrc/exports/movements/models/consolidation/ConsolidationType.scala
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.exports.movements.models.consolidation
+
+import play.api.libs.json.{Format, JsError, JsResult, JsString, JsSuccess, JsValue}
+import uk.gov.hmrc.exports.movements.models.submissions.ActionType
+
+sealed abstract class ConsolidationType(val value: String, val code: String) {
```
ed0906: if you used a Scala Enumeration this file would be about 4 lines long and wouldn't need all the Reads/Writes/all values code (and their tests)

app/uk/gov/hmrc/exports/movements/services/SubmissionService.scala
```diff
@@ -58,6 +56,11 @@ class SubmissionService @Inject()(
         )
     }
 
+  def submitConsolidationRequest(eori: String, consolidationRequest: ConsolidationRequest)(
```
ed0906: If the Service is going to use it can we not call it `Request` please, It suggests Controller concerns are leaking into the Service

conf/app.routes
```diff
@@ -16,3 +16,7 @@ POST        /consolidations/shut                            uk.gov.hmrc.exports.
 POST        /consolidations/associate                       uk.gov.hmrc.exports.movements.controllers.ConsolidationController.associateMucr()
 POST        /consolidations/disassociate                    uk.gov.hmrc.exports.movements.controllers.ConsolidationController.disassociateMucr()
 
+# New Movements Consolidations
+
+POST        /consolidation                                  uk.gov.hmrc.exports.movements.controllers.ConsolidationController.submitConsolidation()
```
ed0906: 👍

app/uk/gov/hmrc/exports/movements/controllers/ConsolidationController.scala
```diff
@@ -81,4 +82,14 @@ class ConsolidationController @Inject()(
           ErrorResponse.errorInternalServerError(e.getMessage).XmlResult
       }
 
+  def submitConsolidation(): Action[ConsolidationRequest] = authorisedAction(parse.json[ConsolidationRequest]) {
+    implicit request =>
+      consolidationService
+        .submitConsolidationRequest(request.eori.value, request.body)
+        .map(_ => Accepted("Consolidation request submitted successfully"))
```
ed0906: Generally `POST` requests to rest endpoints return the "thing that was created". 
I'd return the Consolidation Payload

app/uk/gov/hmrc/exports/movements/models/consolidation/ConsolidationRequest.scala
```diff
@@ -0,0 +1,69 @@
...
+
+import scala.xml.{Node, NodeSeq}
+
+class ConsolidationRequest(
+  val consolidationType: ConsolidationType,
+  val mucr: Option[String],
+  val ducr: Option[String]
+) {
+
+  def generateSubmissionRequestContext(eori: String): SubmissionRequestContext =
```
ed0906: Not sure this really belongs in the `Request` object (for separation of concerns). Would it not make more sense for it to exist in a `WCOService` or something? (similar to the Declaration Back End)

PatrykRudnicki: Moved to the separate service

app/uk/gov/hmrc/exports/movements/models/consolidation/ConsolidationRequest.scala
PatrykRudnicki: Implementation changed

app/uk/gov/hmrc/exports/movements/models/consolidation/ConsolidationType.scala
PatrykRudnicki: With enum you can't specify more information with some value. Every consolidation type contains value that give a possibility to map this from request from frontend, but moreover we need to know what action type for submission is it and code for consolidation xml. Of course we can achieve this with additional methods, but I'm not sure that with all those methods the code base will be shorter or better. I'm open to change it if Scala Enumeration here has some value

PatrykRudnicki:
test this please

app/uk/gov/hmrc/exports/movements/models/consolidation/ConsolidationType.scala
ed0906: You can just do:
```
object SomeType extends Enumeration {
  type SomeType = Value
  val TYPE_1 = Value("Some value")
}
```

app/uk/gov/hmrc/exports/movements/controllers/ConsolidationController.scala
```diff
@@ -81,4 +83,10 @@ class ConsolidationController @Inject()(
           ErrorResponse.errorInternalServerError(e.getMessage).XmlResult
       }
 
+  def submitConsolidation(): Action[ConsolidationRequest] = authorisedAction(parse.json[ConsolidationRequest]) {
+    implicit request =>
+      consolidationService
+        .submitConsolidation(request.eori.value, request.body)
+        .map(_ => Accepted(Json.toJson(request.body)))
```
ed0906: You could pull this across from `customs-declare-exports`

```
trait JSONResponses {

  implicit def writable[T](implicit writes: Writes[T], code: Codec): Writeable[T] = {
    implicit val contentType: ContentTypeOf[T] = ContentTypeOf[T](Some(ContentTypes.JSON))
    Writeable(Writeable.writeableOf_JsValue.transform.compose(writes.writes))
  }
}
```

And just do `Accepted(request.body)`

ed0906: Now we are introducing JSON endpoints we have an opportunity to set the pattern

app/uk/gov/hmrc/exports/movements/models/consolidation/ConsolidationRequest.scala
ed0906: Have you pushed? It still looks the same to me?

app/uk/gov/hmrc/exports/movements/services/SubmissionService.scala
```diff
@@ -58,6 +57,14 @@ class SubmissionService @Inject()(
         )
     }
 
+  def submitConsolidation(eori: String, consolidationRequest: ConsolidationRequest)(
```
ed0906: We are still passing something with the word `Request` in it's name into the service layer which seems a little strange to me

app/uk/gov/hmrc/exports/movements/controllers/ConsolidationController.scala
PatrykRudnicki: Added

app/uk/gov/hmrc/exports/movements/models/consolidation/ConsolidationRequest.scala
PatrykRudnicki: yup, this snippet that we're commenting will be always old, but the implementation is new

app/uk/gov/hmrc/exports/movements/services/SubmissionService.scala
PatrykRudnicki: Changed to consolidation, but IMO ConsolidationRequest it's the name of the xml element and this name was better in places like this.

PatrykRudnicki:
Using one endpoint for three types with similar model causes problems with validation. We must use Options there to give possibility to build model with information that we need to have. I'm not sure that approach with one endpoint it's the best in that case. I know that three endpoints will have more lines, but in practice it can be safer and easier to maintain solution. Now, every solution with one model using enum or sealed abstract class must have additional validation to check that all mandatory information are in the request. Enumeration can be connected with one value inside, we need to have more data, because our structure is complicated. Methods that allow us build consolidation xml need to have more arguments than I think it's necessary. When we switch frontend to use new approach and we remove all old code everything should be easier to maintain and read. I still don't see here place for scala enumeration because we will need to have more pattern matches and everything to build and validate correct request. I think it's a good moment to have discussion and decide about one endpoint vs three endpoint approach.

PatrykRudnicki:
We can introduce three models AssociateDucr, DisassosiateDucr and ShutMucr, three endpoint for it, but after first model validation every of this model will have a method to convert this to ConsolidationRequest, based on the ConsolidationRequest we will generate xml etc.

ed0906:
I'd suggest one endpoint and 3 models which all extend a single trait `ConsolidationRequest`

The three models will have only the mandatory fields that they need. That way Play will validate it for you and you don't need any validation at all

app/uk/gov/hmrc/exports/movements/services/WCOMapper.scala
```diff
@@ -0,0 +1,46 @@
...
+
+package uk.gov.hmrc.exports.movements.services
+
+import javax.inject.Singleton
+import uk.gov.hmrc.exports.movements.models.consolidation.Consolidation
+
+import scala.xml.{Node, NodeSeq}
+
+@Singleton
+class WCOMapper {
```
MaciejRewera: This is not mapping into `wco` domain so it is better to avoid it in the name.

app/uk/gov/hmrc/exports/movements/models/consolidation/Consolidation.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package uk.gov.hmrc.exports.movements.models.consolidation
+
+import play.api.libs.json.Json
+import uk.gov.hmrc.exports.movements.services.context.SubmissionRequestContext
+
+import scala.xml.Node
+
+class Consolidation(val consolidationType: ConsolidationType, val mucr: Option[String], val ducr: Option[String]) {
```
MaciejRewera: This class requires more attention. For e.g. you potetntially can have multiple DUCRs for your consolidation.
I think it would be more robust if you made use of `UcrBlock` class (from our repo, not wco-dec) and here you've had a `Seq` of those.

app/uk/gov/hmrc/exports/movements/models/consolidation/ConsolidationType.scala
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+package uk.gov.hmrc.exports.movements.models.consolidation
+
+import play.api.libs.json.{Format, JsError, JsResult, JsString, JsSuccess, JsValue}
+import uk.gov.hmrc.exports.movements.models.submissions.ActionType
+
+sealed abstract class ConsolidationType(val value: String, val actionType: ActionType, val code: String) {
+
+  def apply(mucr: Option[String], ducr: Option[String]): Consolidation
```
MaciejRewera: I don't think this is the appropriate place for `Consolidation` factory method. I'd expect it to return `ConsolidationType` but not `Consolidation`.

PatrykRudnicki:
@MaciejRewera @ed0906 Implementation with enum is done, I'll work on the tests now, but it'll be good to have feedback earlier

app/uk/gov/hmrc/exports/movements/services/WCOMapper.scala
ed0906: Maybe ILEMapper instead?

app/uk/gov/hmrc/exports/movements/models/consolidation/Consolidation.scala
ed0906: Is it current functionality to allow multiple DUCRs or expected future functionality @MaciejRewera ?

app/uk/gov/hmrc/exports/movements/services/WCOMapper.scala
PatrykRudnicki: Changed to ILEMapper

app/uk/gov/hmrc/exports/movements/models/consolidation/Consolidation.scala
PatrykRudnicki: This class has different implementation right now, but question from @ed0906 still need to be answered

app/uk/gov/hmrc/exports/movements/models/consolidation/ConsolidationType.scala
PatrykRudnicki: This disapper

app/uk/gov/hmrc/exports/movements/models/consolidation/Consolidation.scala
```diff
@@ -0,0 +1,84 @@
...
+    .and[DisassiociateDucrRequest](typeTag = "disassociateDucr")
+    .and[ShutMucrRequest](typeTag = "shutMucr")
+    .format
+}
+
+case class Consolidation(
+  consolidationType: ConsolidationType,
+  mucr: Option[String],
+  ducr: Option[String],
+  actionType: ActionType
```
SteveSugden: Do we need the actionType (if we have the consolidationType)?

app/uk/gov/hmrc/exports/movements/models/consolidation/Consolidation.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  override def consolidation(): Consolidation = Consolidation(SHUT_MUCR, Some(mucr), None, ShutMucr)
+}
+
+object ConsolidationRequest {
+  implicit val associateDucrFormat = Json.format[AssociateDucrRequest]
+  implicit val disassiociateDucrFormat = Json.format[DisassiociateDucrRequest]
+  implicit val shutMucrFormat = Json.format[ShutMucrRequest]
+
+  implicit val format = Union
```
SteveSugden: What does the Union do?  Looks interesting.

ed0906: We had this in Tariffs Steve, it creates JSON formatters for inheritance hierachies

app/uk/gov/hmrc/exports/movements/models/consolidation/Consolidation.scala
ed0906: Hmm yes, why is there two Type fields?

PatrykRudnicki: Action Type is necessary to build submissionRequestContext, I'm not able to get rid of this or change implementation because old endpoint use it. When I remove old endpoint after frontend merge I'll be able to clear this

app/uk/gov/hmrc/exports/movements/models/consolidation/Consolidation.scala
```diff
@@ -0,0 +1,84 @@
...
+  val ASSOCIATE_DUCR = Value("EAC")
+  val DISASSOCIATE_DUCR = Value("EAC")
+  val SHUT_MUCR = Value("CST")
+
+  implicit val format = Format(Reads.enumNameReads(ConsolidationType), Writes.enumNameWrites)
+}
+
+import ConsolidationType._
+
+sealed trait ConsolidationRequest {
```
MaciejRewera: This seems like overengineering to me.
You could make `ConsolidationRequest` an abstract sealed class and have all fields in it. I don't see any reason to use separate `Consolidation` class for this.

app/uk/gov/hmrc/exports/movements/models/consolidation/Consolidation.scala
MaciejRewera: I think we could move it into `ConsolidationType` maybe?

app/uk/gov/hmrc/exports/movements/models/consolidation/Consolidation.scala
PauloPortugal: Another political statement... let's not waste time here, we already have people on parliament sorting it out

app/uk/gov/hmrc/exports/movements/models/consolidation/Consolidation.scala
```diff
@@ -0,0 +1,84 @@
...
+
+import play.api.libs.json.{Format, Json, Reads, Writes}
+import uk.gov.hmrc.exports.movements.models.submissions.ActionType
+import uk.gov.hmrc.exports.movements.models.submissions.ActionType.{DucrAssociation, DucrDisassociation, ShutMucr}
+import uk.gov.hmrc.exports.movements.services.context.SubmissionRequestContext
+import uk.gov.hmrc.play.json.Union
+
+import scala.xml.Node
+
+object ConsolidationType extends Enumeration {
```
MaciejRewera: It might be better to move it to separate file.

PR #41, opened 2019-10-04
CEDS-1766 Move Arrive/Depart submission to the backend

PauloPortugal:


app/uk/gov/hmrc/exports/movements/controllers/MovementsController.scala
```diff
@@ -0,0 +1,73 @@
...
+)(implicit executionContext: ExecutionContext)
+    extends RESTController(controllerComponents) {
+
+  private val logger = Logger(this.getClass)
+
+  def submitMovement(): Action[MovementRequest] =
+    authenticator.authorisedAction(parsingJson[MovementRequest]) { implicit request =>
+      val data: InventoryLinkingMovementRequest = request.body.createMovementRequest(request.eori)
+      request.body.choice match {
+        case "EAL" =>
```
PatrykRudnicki: IMO it will be good to hide `EAL`, `EDL` somehow, send choice from frontend (movements use new approach for choice) and based on this create value

app/uk/gov/hmrc/exports/movements/controllers/MovementsController.scala
```diff
@@ -0,0 +1,73 @@
...
+    submissionService
+      .submitRequest(
+        SubmissionRequestContext(
+          eori = eori.value,
+          actionType = actionType,
+          requestXml = xml.XML.loadString(data.toXml)
+        )
+      )
+      .map(_ => Accepted("Movement Submission submitted successfully"))
+      .recover {
```
PatrykRudnicki: As I talked with @ed0906 this is unnecessary

app/uk/gov/hmrc/exports/movements/controllers/SubmissionController.scala
```diff
@@ -90,6 +86,11 @@ class SubmissionController @Inject()(
           ErrorResponse.errorInternalServerError(e.getMessage).XmlResult
       }
 
+  def submitDeparture(): Action[AnyContent] =
+    authorisedAction(bodyParser = xmlOrEmptyBody) { implicit request =>
```
PatrykRudnicki: We should send json from frontend

app/uk/gov/hmrc/exports/movements/controllers/request/MovementRequest.scala
```diff
@@ -0,0 +1,71 @@
...
+) {
+  def createMovementRequest(eori: Eori): InventoryLinkingMovementRequest = {
+
+    val departureDetails = choice match {
+      case "EDL" => Some(movementDetails)
+      case _     => None
+    }
+
+    val arrivalDetails: Option[MovementDetails] = choice match {
+      case "EAL" => Some(movementDetails)
```
PatrykRudnicki: The same comment like above. I'll hide those value somehow

app/uk/gov/hmrc/exports/movements/models/ErrorResponse.scala
```diff
@@ -40,17 +40,13 @@ object ResponseContents {
 case class ErrorResponse(httpStatusCode: Int, errorCode: String, message: String, content: ResponseContents*)
     extends Error {
 
+  lazy val JsonResult: Result = Status(httpStatusCode)(responseJson).as(ContentTypes.JSON)
+  lazy val XmlResult: Result = Status(httpStatusCode)(responseXml).as(ContentTypes.XML)
```
PatrykRudnicki: I'm not sure that there is some value of using lazy in whole ErrorResponse

app/uk/gov/hmrc/exports/movements/controllers/MovementsController.scala
```diff
@@ -0,0 +1,73 @@
...
+        case e: CustomsInventoryLinkingUpstreamException =>
+          ErrorResponse.errorInternalServerError(e.getMessage).XmlResult
```
bambuchaAdm: why XML as response in new endpoint ?

app/uk/gov/hmrc/exports/movements/controllers/RESTController.scala
```diff
@@ -0,0 +1,44 @@
...
+      case JsError(errors) =>
+        val payload = Json.toJson(ErrorResponse("Bad Request", Some(errors.map {
+          case (path, errs) => path.toString() + ": " + errs.map(_.message).headOption.getOrElse("unknown")
+        })))
+        logger.warn(s"Bad Request [$payload]")
+        Left(BadRequest(payload))
+    }
+  }
+
+}
```
bambuchaAdm: AFAIR here we would have duplication with @PatrykRudnicki

app/uk/gov/hmrc/exports/movements/controllers/actions/Authenticator.scala
```diff
@@ -0,0 +1,80 @@
...
+import uk.gov.hmrc.http.HeaderCarrier
+import uk.gov.hmrc.play.bootstrap.controller.BackendController
+
+import scala.concurrent.{ExecutionContext, Future}
+
+// TODO: This needs to be separated from BackendController and be injected instead of inheriting in other controllers
+@Singleton
+class Authenticator @Inject()(override val authConnector: AuthConnector, cc: ControllerComponents)(
+  implicit ec: ExecutionContext
+) extends BackendController(cc) with AuthorisedFunctions {
```
bambuchaAdm: You can inject Authorization or something similar

app/uk/gov/hmrc/exports/movements/controllers/response/ErrorResponse.scala
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.exports.movements.controllers.response
+
+import play.api.libs.json.{Json, OFormat}
+
+case class ErrorResponse(message: String, errors: Option[Seq[String]] = None)
```
bambuchaAdm: Why we have this ? Is JsonErrorHandler from bootstrap-play-26 is not enough ?

app/uk/gov/hmrc/exports/movements/models/AuthorisedRequest.scala
```diff
@@ -27,6 +28,10 @@ trait HasConversationId {
 }
 
 case class Eori(value: String) extends AnyVal
+object Eori {
+  implicit val format: OFormat[Eori] = Json.format[Eori]
```
bambuchaAdm: Please use `Reads.StringReads.transform(???)` - now outcome of json would be
```
{
  "eori" : { "value" : "GB1245324235235" }
}
```

app/uk/gov/hmrc/exports/movements/models/ErrorResponse.scala
```diff
@@ -64,17 +60,18 @@ object ErrorResponse extends HttpStatusCodeShortDescriptions {
...
+  def errorInvalidPayload: ErrorResponse = errorBadRequest("Invalid payload")
 
-  val ErrorGenericBadRequest: ErrorResponse = errorBadRequest("Bad Request")
+  def errorInvalidHeaders: ErrorResponse = errorBadRequest("Invalid headers")
 
-  val ErrorInvalidPayload: ErrorResponse = errorBadRequest("Invalid payload")
+  def errorBadRequest(errorMessage: String): ErrorResponse = ErrorResponse(BAD_REQUEST, BadRequestCode, errorMessage)
 
-  val ErrorInternalServerError: ErrorResponse = errorInternalServerError("Internal server error")
+  def errorInternalServerError: ErrorResponse = errorInternalServerError("Internal server error")
```
bambuchaAdm: If we move to usage more JsonErrorHanlder this class would be completely in required

app/uk/gov/hmrc/exports/movements/controllers/MovementsController.scala
```diff
@@ -0,0 +1,73 @@
...
+  submissionService: SubmissionService,
+  override val controllerComponents: ControllerComponents
+)(implicit executionContext: ExecutionContext)
+    extends RESTController(controllerComponents) {
+
+  private val logger = Logger(this.getClass)
+
+  def submitMovement(): Action[MovementRequest] =
+    authenticator.authorisedAction(parsingJson[MovementRequest]) { implicit request =>
+      val data: InventoryLinkingMovementRequest = request.body.createMovementRequest(request.eori)
```
ed0906: I don’t think the controller should care about ILE, sounds like a service level concern to me

app/uk/gov/hmrc/exports/movements/models/movements/Transport.scala
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.exports.movements.models.movements
+import play.api.libs.json.{Json, OFormat}
+
+case class Transport(modeOfTransport: String, nationality: String, transportId: String)
```
bambuchaAdm: Can we put all of those classes into single file - it is not Java wolds any more XD

app/uk/gov/hmrc/exports/movements/controllers/MovementsController.scala
ed0906: Also, POST endpoints usually return the “thing that was created”. I’d just return the movement request

app/uk/gov/hmrc/exports/movements/controllers/RESTController.scala
PatrykRudnicki: I added JSONResponses, that @PauloPortugal extends here so it's okay :)

app/uk/gov/hmrc/exports/movements/controllers/actions/Authenticator.scala
```diff
@@ -0,0 +1,80 @@
...
+import uk.gov.hmrc.exports.models.ErrorResponse
+import uk.gov.hmrc.exports.movements.models.{AuthorizedSubmissionRequest, Eori}
+import uk.gov.hmrc.http.HeaderCarrier
+import uk.gov.hmrc.play.bootstrap.controller.BackendController
+
+import scala.concurrent.{ExecutionContext, Future}
+
+// TODO: This needs to be separated from BackendController and be injected instead of inheriting in other controllers
+@Singleton
+class Authenticator @Inject()(override val authConnector: AuthConnector, cc: ControllerComponents)(
```
ed0906: Do we not already have Auth?

app/uk/gov/hmrc/exports/movements/controllers/request/MovementRequest.scala
```diff
@@ -0,0 +1,71 @@
...
+
+package uk.gov.hmrc.exports.movements.controllers.request
+
+import play.api.libs.json.{Json, OFormat}
+import uk.gov.hmrc.exports.movements.models.Eori
+import uk.gov.hmrc.exports.movements.models.movements._
+import uk.gov.hmrc.wco.dec.inventorylinking.common.{TransportDetails, UcrBlock}
+import uk.gov.hmrc.wco.dec.inventorylinking.movement.request.InventoryLinkingMovementRequest
+
+case class MovementRequest(
```
ed0906: How similar are Arrival and Departure requests? Are all the same fields mandatory?

If not we should have two models and use inheritance e.g in Patryks PR

conf/app.routes
```diff
@@ -7,6 +7,8 @@ GET         /notifications/:conversationId                  uk.gov.hmrc.exports.
 POST        /movements/arrival                              uk.gov.hmrc.exports.movements.controllers.SubmissionController.submitArrival()
 POST        /movements/departure                            uk.gov.hmrc.exports.movements.controllers.SubmissionController.submitDeparture()
 
+POST        /movements                                      uk.gov.hmrc.exports.movements.controllers.MovementsController.submitMovement()
```
ed0906: Can we call the controller method createMovement?

ed0906: Keep the /movements though that’s good

app/uk/gov/hmrc/exports/movements/controllers/actions/Authenticator.scala
PauloPortugal: This is not longer needed. Removed.

app/uk/gov/hmrc/exports/movements/models/ErrorResponse.scala
PauloPortugal: I was following the example there. Would you recommend removing the `lazy` keyword

app/uk/gov/hmrc/exports/movements/controllers/SubmissionController.scala
PauloPortugal: This is for the current state.
1. Deploy BE with two endpoints: one JSON and one XML
2. Allow the FE to submit XML to BE, so that smoke tests do not fail
3. Deploy FE to only submit JSON. Smoke test should still be successful 
4. Remove the old XML endpoint

This makes it backwards compatible.

app/uk/gov/hmrc/exports/movements/controllers/MovementsController.scala
PauloPortugal: Good point! Resolved.

app/uk/gov/hmrc/exports/movements/controllers/RESTController.scala
PauloPortugal: This is a pattern that we do not use here. Removed.

conf/app.routes
PauloPortugal: Done.

app/uk/gov/hmrc/exports/movements/controllers/actions/Authenticator.scala
PauloPortugal: REmoved.

app/uk/gov/hmrc/exports/movements/controllers/response/ErrorResponse.scala
PauloPortugal: Good question, probably a tech debt task after this work goes in.

app/uk/gov/hmrc/exports/movements/controllers/MovementsController.scala
PauloPortugal: This has been addressed.

app/uk/gov/hmrc/exports/movements/controllers/request/MovementRequest.scala
PauloPortugal: This has been addressed.

app/uk/gov/hmrc/exports/movements/controllers/request/MovementRequest.scala
PauloPortugal: The mandatory fields are the same

app/uk/gov/hmrc/exports/movements/controllers/MovementsController.scala
```diff
@@ -0,0 +1,67 @@
...
+  authConnector: AuthConnector,
+  submissionService: SubmissionService,
+  override val controllerComponents: ControllerComponents
+)(implicit executionContext: ExecutionContext)
+    extends AuthenticatedController(authConnector, controllerComponents) {
+
+  private val logger = Logger(this.getClass)
+
+  def createMovement(): Action[MovementRequest] = authorisedAction(parse.json[MovementRequest]) { implicit request =>
+    val data: InventoryLinkingMovementRequest = request.body.createMovementRequest(request.eori)
```
PatrykRudnicki: I think this generation InventoryLinkingMovementRequest should be in some factory and whole controller should be based on the json that we received from frontend.

app/uk/gov/hmrc/exports/movements/controllers/MovementsController.scala
```diff
@@ -0,0 +1,67 @@
...
+        submitMovementSubmission(data: InventoryLinkingMovementRequest, request.eori, ActionType.Departure)
+    }
+  }
+
+  private def submitMovementSubmission(data: InventoryLinkingMovementRequest, eori: Eori, actionType: ActionType)(
+    implicit hc: HeaderCarrier
+  ): Future[Result] =
+    submissionService
+      .submitRequest(
+        SubmissionRequestContext(
```
PatrykRudnicki: I get rid of usage in consolidations `SubmissionRequestContext` in open PR now, I think it'll be really good, when we'll not base our submission on this class. Another wrapper that is not useful here, because we can build Submission instead of this and for submission Service just put information that we need to have.

app/uk/gov/hmrc/exports/movements/controllers/MovementsController.scala
```diff
@@ -0,0 +1,67 @@
...
+          eori = eori.value,
+          actionType = actionType,
+          requestXml = xml.XML.loadString(data.toXml)
+        )
+      )
+      .map(_ => Accepted("Movement Submission submitted successfully"))
```
PatrykRudnicki: As I discussed with @ed0906 before about consolidations we should return to frontend this request model.

app/uk/gov/hmrc/exports/movements/controllers/request/MovementRequest.scala
```diff
@@ -0,0 +1,72 @@
...
+  choice: String,
+  consignmentReference: ConsignmentReference,
+  movementDetails: MovementDetails,
+  location: Option[Location] = None,
+  arrivalReference: Option[ArrivalReference] = None,
+  goodsDeparted: Option[GoodsDeparted] = None,
+  transport: Option[Transport] = None
+) {
+  def createMovementRequest(eori: Eori): InventoryLinkingMovementRequest = {
```
PatrykRudnicki: I'll do this in ILEMapper, but it can be done in separate PR to clear stuff when migration to the backend will be done

app/uk/gov/hmrc/exports/movements/controllers/MovementsController.scala
PauloPortugal: I have addressed this

PR #47, opened 2019-10-18
CEDS-1812-disassociate-mucr

SteveSugden:
This PR is to extend the 'Consolidation' message to handle two new MUCR-MUCR types so that CEDS-1595 and CEDS-1812 can be worked on in parallel.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,6 +7,7 @@ import uk.gov.hmrc.{SbtArtifactory, SbtAutoBuildPlugin}
 
 val appName = "customs-declare-exports-movements"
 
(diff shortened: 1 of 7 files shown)


build.sbt
```diff
@@ -7,6 +7,7 @@ import uk.gov.hmrc.{SbtArtifactory, SbtAutoBuildPlugin}
 
 val appName = "customs-declare-exports-movements"
 
+PlayKeys.devSettings := Seq("play.server.http.port" -> "6797")
```
ed0906: 👍

SteveSugden: That was an accidental commit.  It's been hanging around on my machine for ages.

PR #50, opened 2019-10-22
CEDS-1858 Switch Client ID depending on User Agent

ed0906:
In MDTP environments requests always send with a `User-Agent` header
We can use this to switch between the Movements & Internal UI Client IDs (they both need their own)

Annoyingly when running locally this header isn't sent so there is a fallback which can be blanked out in non-local environments


ed0906:
test this please

test/unit/uk/gov/hmrc/exports/movements/connector/CustomsInventoryLinkingExportsConnectorSpec.scala
```diff
@@ -16,61 +16,52 @@
...
-import uk.gov.hmrc.http.logging.Authorization
-import unit.uk.gov.hmrc.exports.movements.base.{CustomsExportsBaseSpec, MockHttpClient}
 
-import scala.concurrent.Future
-import scala.util.Random
-import scala.xml.{Elem, NodeSeq}
+import scala.xml.Elem
 
-class CustomsInventoryLinkingExportsConnectorSpec extends CustomsExportsBaseSpec {
+class CustomsInventoryLinkingExportsConnectorSpec extends ConnectorSpec {
```
PatrykRudnicki: If this test is integration test and ConnectorSpec contains application please move it outside package `unit`

app/uk/gov/hmrc/exports/movements/config/AppConfig.scala
```diff
@@ -32,10 +35,16 @@ class AppConfig @Inject()(runModeConfiguration: Configuration, servicesConfig: S
     throw new IllegalStateException("Missing configuration for Customs Inventory Linking send arrival URI")
   )
 
-  lazy val clientIdInventory = servicesConfig.getConfString(
-    "customs-inventory-linking-exports.client-id",
-    throw new IllegalStateException("Missing configuration for Customs Inventory Linking Client Id")
-  )
+  def clientIdInventory(implicit hc: HeaderCarrier): String = {
```
bambuchaAdm: Can we move this as single loaded `Map[String, String]` with default value ?

conf/application.conf
```diff
@@ -160,7 +160,9 @@ microservice {
             host = localhost
             port = 9823
             sendArrival = "/"
-            client-id = "5c68d3b5-d8a7-4212-8688-6b67f18bbce7"
+            client-id {
+                default = "5c68d3b5-d8a7-4212-8688-6b67f18bbce7"
```
bambuchaAdm: What if received client-id would be default ?

ed0906: Not sure I understand

app/uk/gov/hmrc/exports/movements/config/AppConfig.scala
ed0906: I see no helper methods for getting Maps out of config, do you know of one?

PR #54, opened 2019-10-24
CEDS - Add movements frontend client ID for local use

ed0906:


conf/application.conf
```diff
@@ -162,8 +162,11 @@ microservice {
...
+                customs-movements-frontend = "5c68d3b5-d8a7-4212-8688-6b67f18bbce7"
             }
             schema-file-path = "conf/schemas/exports/inventoryLinkingResponseExternal.xsd"
         }
     }
-}
\ No newline at end of file
+}
+
+httpHeadersWhitelist = ["User-Agent"]
```
MaciejRewera: shouldn't this be lower-cased ?

PatrykRudnicki: It can be. First probably it works as no lower-case and second Ed converts this to lover case in the appconfig

conf/application.conf
```diff
@@ -162,8 +162,11 @@ microservice {
             sendArrival = "/"
             client-id {
                 default = "5c68d3b5-d8a7-4212-8688-6b67f18bbce7"
+                customs-movements-frontend = "5c68d3b5-d8a7-4212-8688-6b67f18bbce7"
```
PatrykRudnicki: I think we should update environments. To have a correct path

conf/application.conf
ed0906: This isnt part of this PR, I just havent pulled down on this branch

ed0906: Headers are case insensitive, I've updated my branch anyway, this shouldnt have been part of it

conf/application.conf
ed0906: Not sure what you mean?

PatrykRudnicki: Nevermind, If we're taking application conf to all environments it should works, I can't remember that not overridden settings from application.conf are used on the environments

PR #55, opened 2019-10-25
Add 3 new CHIEF errors

MaciejRewera:
Some time ago we have received these error codes and they were added to
the Frontend. Now it is time to add it to the Backend.

### DIFF
--- a/conf/chief_errors.csv
+++ b/conf/chief_errors.csv
@@ -23,3 +23,6 @@ E607,Entry Reference is not suitable for Export movement
 E3464,Refused – User Role not matched on MUCR arrivals
 E3480,Only shut MUCRs can be associated into another MUCR
 E1264,Cannot shut the consolidation since it is moving
+E518,	Shed Operator does not exist
+E10334, Service segment invalid
+E10003, Errors on Document


conf/chief_errors.csv
```diff
@@ -22,4 +22,7 @@ E3479,Unable to depart MUCR as it is contained in another MUCR
 E607,Entry Reference is not suitable for Export movement
 E3464,Refused – User Role not matched on MUCR arrivals
 E3480,Only shut MUCRs can be associated into another MUCR
-E1264,Cannot shut the consolidation since it is moving
\ No newline at end of file
+E1264,Cannot shut the consolidation since it is moving
+E518,	Shed Operator does not exist
+E10334, Service segment invalid	Service segment invalid
+E10003, Errors on Document	Errors on Document
```
PatrykRudnicki: This huge space between `invalid` and `Service` is something that you want to have?

The same with `Document` and `Errors` in the next line

MaciejRewera: Nah, this is an error. Though there can't be a problem with such a small PR...

PR #59, opened 2019-11-04
CEDS Add /notifications?eori=x&pid=y endpoint

ed0906:


test/unit/uk/gov/hmrc/exports/movements/controllers/NotificationControllerSpecNoApp.scala
```diff
@@ -55,14 +56,17 @@ class NotificationControllerSpecNoApp extends WordSpec with MustMatchers with Mo
     implicit val materializer: ActorMaterializer = ActorMaterializer()
   }
 
-  "Notification Controller" should {
+  "GET notifications/:action-id" should {
```
SteveSugden: What's the story with "actionId" v.s. "conversationId"?  Can these be made the same?

PR #60, opened 2019-11-05
CEDS Fix logic behind GET /notifications endpoint

MaciejRewera:


app/uk/gov/hmrc/exports/movements/services/NotificationService.scala
```diff
@@ -38,18 +37,13 @@ class NotificationService @Inject()(notificationRepository: NotificationReposito
 
   def getAllNotifications(searchParameters: SearchParameters): Future[Seq[NotificationFrontendModel]] =
     submissionRepository.findBy(searchParameters).flatMap {
-      case Nil => returnFailed()
-      case _   => returnSuccess(searchParameters)
+      case Nil         => Future.successful(Seq.empty)
```
ed0906: Will it ever be `Nil`? I'd imagine it would just be an empty sequence if none were found

app/uk/gov/hmrc/exports/movements/repositories/NotificationRepository.scala
```diff
@@ -36,7 +36,10 @@ class NotificationRepository @Inject()(mc: ReactiveMongoComponent)(implicit ec:
     Index(Seq("conversationId" -> IndexType.Ascending), name = Some("conversationIdIdx"))
   )
 
-  def findByConversationId(conversationId: String): Future[Seq[Notification]] =
-    find("conversationId" -> JsString(conversationId))
+  def findByConversationIds(conversationIds: Seq[String]): Future[Seq[Notification]] =
+    conversationIds match {
```
PatrykRudnicki: Are you sure that `find`can't handle `Seq.empty` automatically?

app/uk/gov/hmrc/exports/movements/controllers/NotificationController.scala
```diff
@@ -78,10 +78,10 @@ class NotificationController @Inject()(
         metrics.incrementCounter(movementMetric)
     }
 
-  def getNotificationsForAction(eori: Option[String], providerId: Option[String], conversationId: String): Action[AnyContent] =
+  def getNotificationsForSubmission(eori: Option[String], providerId: Option[String], conversationId: String): Action[AnyContent] =
     searchNotifications(eori, providerId, Some(conversationId))
 
-  def getNotifications(eori: Option[String], providerId: Option[String]): Action[AnyContent] =
+  def getAllNotificationsForUser(eori: Option[String], providerId: Option[String]): Action[AnyContent] =
```
PauloPortugal: :+1: Clearer naming is always good! :)

conf/app.routes
```diff
@@ -2,8 +2,8 @@
 
 POST        /customs-declare-exports/notifyMovement         uk.gov.hmrc.exports.movements.controllers.NotificationController.saveNotification()
 
-GET         /notifications                                  uk.gov.hmrc.exports.movements.controllers.NotificationController.getNotifications(eori: Option[String], providerId: Option[String])
-GET         /notifications/:conversationId                  uk.gov.hmrc.exports.movements.controllers.NotificationController.getNotificationsForAction(eori: Option[String], providerId: Option[String], conversationId: String)
+GET         /notifications                                  uk.gov.hmrc.exports.movements.controllers.NotificationController.getAllNotificationsForUser(eori: Option[String], providerId: Option[String])
```
ed0906: I think `getNotifications` made more sense as both query params are optional

app/uk/gov/hmrc/exports/movements/services/NotificationService.scala
```diff
@@ -38,18 +37,13 @@ class NotificationService @Inject()(notificationRepository: NotificationReposito
...
+      case submissions => getNotifications(submissions.map(_.conversationId))
```
PatrykRudnicki: The same here.
```
submissionRepository.findBy(searchParameters).flatMap { submissions =>
   getNotifications(submissions.map(_.conversationId))
}
```

test/unit/uk/gov/hmrc/exports/movements/services/NotificationServiceSpec.scala
```diff
@@ -124,13 +124,79 @@ class NotificationServiceSpec extends WordSpec with MockitoSugar with ScalaFutur
...
+
+        val submissions = Seq(exampleSubmission(), exampleSubmission(conversationId = conversationId_2))
+        when(submissionRepositoryMock.findBy(meq(searchParameters))).thenReturn(Future.successful(submissions))
+
+        notificationService.getAllNotifications(searchParameters).futureValue
+
+        val captor: ArgumentCaptor[Seq[String]] = ArgumentCaptor.forClass(classOf[Seq[String]])
+        verify(notificationRepositoryMock).findByConversationIds(captor.capture())
+
+        captor.getValue mustBe Seq(conversationId, conversationId_2)
```
ed0906: When using `ArgumentCaptor`s can we wrap them in a method, its not the prettiest code

e.g.
```
def theConversationIds: Seq[String] = {
    ...
}
```

test/unit/uk/gov/hmrc/exports/movements/services/NotificationServiceSpec.scala
```diff
@@ -124,13 +124,79 @@ class NotificationServiceSpec extends WordSpec with MockitoSugar with ScalaFutur
         returnedNotifications must contain(expectedSecondNotificationPresentationData)
       }
 
-      "return empty list, if repository returns empty list" in new Test {
+      "return empty list, if NotificationRepository returns empty list" in new Test {
```
ed0906: Not sure we really need this test to be honest. The the test above covers the same scenario more or less

test/unit/uk/gov/hmrc/exports/movements/services/NotificationServiceSpec.scala
```diff
@@ -124,13 +124,79 @@ class NotificationServiceSpec extends WordSpec with MockitoSugar with ScalaFutur
...
+        val returnedNotifications = notificationService.getAllNotifications(searchParameters).futureValue
+
+        val expectedNotifications = notifications.map(NotificationFrontendModel(_))
+        returnedNotifications.length must equal(expectedNotifications.length)
+        expectedNotifications.foreach { notification =>
+          returnedNotifications must contain(notification)
+        }
+      }
+
+      "return empty list, if NotificationRepository returns empty list" in new Test {
```
ed0906: I don't think we need to specific tests for "notification repository returns x", "notification repository returns y"

It's basically the same test. We just care that the repository response is returned from the service

app/uk/gov/hmrc/exports/movements/repositories/NotificationRepository.scala
ed0906: I'd argue its the repositories job to optimise for this case if required

app/uk/gov/hmrc/exports/movements/services/NotificationService.scala
ed0906: Agreed, I think the repository should handle any optimisation for this kind of case

app/uk/gov/hmrc/exports/movements/repositories/NotificationRepository.scala
MaciejRewera: It is not that it won't handle. The result will be the same, but here it is just a minor optimisation not to call DB when we have nothing to ask for.

app/uk/gov/hmrc/exports/movements/services/NotificationService.scala
MaciejRewera: It is eventually the same. The only difference is that `Nil` is a `List`, but in our case it is all the same.

conf/app.routes
MaciejRewera: The endpoint's name stays the same. The methods names are for us, so it is more clear what you could expeect from the meethod.

PR #62, opened 2019-11-13
CEDS-1969 Send Extra auth headers to ILE for Internal Requests

ed0906:
This:
- Sends extra headers `X-Badge-Identifier` and `X-Submitter-Identifier` to the ILE API for internal submissions. Note we send fake data for these for now.
- Removes the `X-EORI-Identifier` as the ILE API doesn't use it
- Removes the `Request` suffix from `MovementRequest` and `ConsolidationRequest` as it added unnecessary clutter and polluted the whole stack with "Request" which is a Controller layer concern.
- Combines the `ILEMapper` and `WCOMapper`. They are both `ILEMapper`s really, one does `Consolidation`s, one does `Movement`s.


app/uk/gov/hmrc/exports/movements/connectors/CustomsInventoryLinkingExportsConnector.scala
```diff
@@ -21,10 +21,10 @@ import play.api.Logger
 import play.api.http.{ContentTypes, HeaderNames}
 import play.api.mvc.Codec
 import play.mvc.Http.Status
-import uk.gov.hmrc.exports.movements.config.AppConfig
+import uk.gov.hmrc.exports.movements.config.{AppConfig}
```
PatrykRudnicki: Unnecessary parenthesis in import

app/uk/gov/hmrc/exports/movements/connectors/CustomsInventoryLinkingExportsConnector.scala
```diff
@@ -34,33 +34,32 @@ import scala.xml.NodeSeq
...
-      HeaderNames.ACCEPT -> "application/vnd.hmrc.1.0+xml",
-      HeaderNames.CONTENT_TYPE -> ContentTypes.XML(Codec.utf_8),
-      CustomsHeaderNames.XClientIdName -> appConfig.clientIdInventory,
-      CustomsHeaderNames.XEoriIdentifierHeaderName -> eori
-    )
+  private def headers(identification: UserIdentification)(implicit hc: HeaderCarrier): Seq[(String, String)] = {
+    val authHeaders = identification.providerId match {
+      case Some(_) =>
+        Seq(
+          CustomsHeaderNames.BadgeIdentifier -> "ABC123",
```
SteveSugden: Is 'ABC123' a placeholder for some later value?

app/uk/gov/hmrc/exports/movements/connectors/CustomsInventoryLinkingExportsConnector.scala
```diff
@@ -34,33 +34,32 @@ import scala.xml.NodeSeq
...
+          CustomsHeaderNames.SubmitterIdentifier -> "ABC123",
```
ed0906: As far as we are aware nothing validates these. We just need to send these headers for ILE APIs validation

app/uk/gov/hmrc/exports/movements/connectors/CustomsInventoryLinkingExportsConnector.scala
```diff
@@ -34,33 +34,32 @@ import scala.xml.NodeSeq
...
+          CustomsHeaderNames.XClientIdName -> appConfig.clientIdInventory
+        )
+      case _ => Seq(CustomsHeaderNames.XClientIdName -> appConfig.clientIdInventory)
+    }
```
PatrykRudnicki: I think that in situation where you don't need to have access to data there is no point of using pattern matching, simple `if` with` isDefined` can be better, but it's just personal opinion

test/integration/uk/gov/hmrc/exports/movements/connector/WiremockTestServer.scala
```diff
@@ -31,14 +31,15 @@ trait WiremockTestServer extends WordSpec with BeforeAndAfterAll {
...
-    super.beforeAll()
+  override protected def beforeEach(): Unit = {
+    super.beforeEach()
     wireMockServer.start()
     WireMock.configureFor(wireHost, wirePort)
   }
 
-  override protected def afterAll(): Unit = {
+  override protected def afterEach(): Unit = {
+    wireMockServer.resetAll()
```
PatrykRudnicki: Is it necessary to reset the `wireMockServer` if we're stopping it in next line?

app/uk/gov/hmrc/exports/movements/services/WCOMapper.scala
```diff
@@ -62,4 +71,34 @@ class WCOMapper {
           transportNationality = Some(data.nationality)
       )
     )
+
+  def generateConsolidationXml(consolidation: Consolidation): Node =
```
PatrykRudnicki: I'm not sure that WCOMapper is a good place for it. It's not related with WCODec library

app/uk/gov/hmrc/exports/movements/connectors/CustomsInventoryLinkingExportsConnector.scala
ed0906: Suprised Scalafmt doesnt handle that

app/uk/gov/hmrc/exports/movements/services/WCOMapper.scala
ed0906: I'll rename it ILEMapper

app/uk/gov/hmrc/exports/movements/connectors/CustomsInventoryLinkingExportsConnector.scala
ed0906: Nope, apparently we just need to populate the header, doesnt matter what with aslong as it meets their regex for that field. Conveniently `ABC123` matches both :)

PR #66, opened 2019-12-05
CEDS-1915 Remove millis from payload datetime element in Movements

MaciejRewera:


test/unit/uk/gov/hmrc/exports/movements/services/ILEMapperSpec.scala
```diff
@@ -38,13 +43,16 @@ class ILEMapperSpec extends UnitSpec {
...
         val input = exampleRetrospectiveArrivalRequest
 
         val xml = ileMapper.generateInventoryLinkingMovementRequestXml(input)
 
-        assert((xml \ "goodsArrivalDateTime").nonEmpty)
+        val dateTimeNode = xml \ "goodsArrivalDateTime"
+        dateTimeNode shouldNot be(empty)
+        val format = DateTimeFormat.forPattern("yyyy-MM-dd'T'HH:mm:ssZ")
+        format.parseDateTime(dateTimeNode.text)
       }
```
SteveSugden: Is this missing an assertion (that the datetime format is correct)?

app/uk/gov/hmrc/exports/movements/services/ILEMapper.scala
```diff
@@ -16,34 +16,38 @@
...
+    val departureDetails: Option[String] = request.choice match {
+      case Departure => request.movementDetails.map(movement => formatOutputDateTime(parseDateTime(movement.dateTime)))
       case _         => None
     }
 
-    val arrivalDetails: Option[MovementDetails] = request.choice match {
-      case Arrival              => request.movementDetails
-      case RetrospectiveArrival => Some(MovementDetails(DateTimeFormatter.ISO_INSTANT.format(Instant.now())))
+    val arrivalDetails: Option[String] = request.choice match {
+      case Arrival              => request.movementDetails.map(movement => formatOutputDateTime(parseDateTime(movement.dateTime)))
```
PatrykRudnicki: I think it's worth to move this method to model to use just `movement.dateTime.format()`

test/integration/uk/gov/hmrc/exports/movements/connector/ConnectorSpec.scala
```diff
@@ -16,25 +16,30 @@
...
 class ConnectorSpec extends WordSpec with GuiceOneAppPerSuite with WiremockTestServer with MustMatchers with MockitoSugar with BeforeAndAfterEach {
 
-  def overrideConfig: Map[String, Any] =
-    Map()
+  private val clock = Clock.fixed(Instant.parse(dateTimeString), ZoneOffset.UTC)
 
   override def fakeApplication(): Application = {
     SharedMetricRegistries.clear()
-    new GuiceApplicationBuilder().configure(overrideConfig).build()
+    new GuiceApplicationBuilder()
```
PatrykRudnicki: What about `overrideConfig` from previous implementation. Is it not necessary?

test/unit/uk/gov/hmrc/exports/movements/services/ILEMapperSpec.scala
```diff
@@ -16,16 +16,20 @@
...
+import org.scalatestplus.mockito.MockitoSugar
 import uk.gov.hmrc.exports.movements.models.consolidation.Consolidation.AssociateDucrRequest
 import uk.gov.hmrc.exports.movements.services.ILEMapper
 import unit.uk.gov.hmrc.exports.movements.base.UnitSpec
 import utils.testdata.CommonTestData._
 import utils.testdata.ConsolidationTestData.exampleAssociateDucrConsolidationRequestXML
 import utils.testdata.MovementsTestData._
 
-class ILEMapperSpec extends UnitSpec {
+class ILEMapperSpec extends UnitSpec with MockitoSugar {
```
PatrykRudnicki: You're not using `MockitoSugar` in this spec

test/integration/uk/gov/hmrc/exports/movements/connector/ConnectorSpec.scala
MaciejRewera: The `overrideConfig` was equal to empty Map. So it was not changing anything.

MaciejRewera: The overrideConfig was equal to empty Map. So it was not changing anything.

app/uk/gov/hmrc/exports/movements/services/ILEMapper.scala
MaciejRewera: I agree that it would make this statement shorter.
But from responsibility perspective `Movement` class serves as a simple DTO.
From a different angle - `ILEMapper` is the place where we perform all the transformations we need.
I'd rather have this line shorter, but sometimes you have to live with it.

PatrykRudnicki: you can make `Option[MovementDetails]` as argument to this class and hide logic inside. The return type is still Option, so only `formatOutputDateTime` will have different signature

`formatOutputDateTime(request.movementsDetails)`

```
private def formatOutputDateTime(movementDetails: MovementDetails): Option[String] = 
movementDetails.map( details => dateTimeFormatter.format(details.dateTime.truncatedTo(ChronoUnit.SECONDS))
```

PatrykRudnicki: But it's okay now :) Just ideas

MaciejRewera: I was thinking about the same solution.
The reason I decided not to do so was that this way you are "bending" the method just for the sake of having shorter statement somewhere else. In this case I decided to make the method as simple and flexible as possible. It should not be tailored for a single use case.

PR #69, opened 2019-12-06
ILE schemas update

MaciejRewera:


MaciejRewera:
New location for xsd files was incorrect. Moved back to the previous directory.

PR #70, opened 2019-12-16
Dependencies update

PatrykRudnicki:


project/plugins.sbt
```diff
@@ -27,3 +27,5 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-uglify" % "1.0.4")
 addSbtPlugin("com.typesafe.sbt" % "sbt-digest" % "1.1.2")
 
 addSbtPlugin("com.lucidchart" % "sbt-scalafmt" % "1.16")
+
+addSbtPlugin("uk.gov.hmrc" % "sbt-git-stamp" % "6.1.0")
```
PatrykRudnicki: We're using gitStampSettings in build.sbt with upgrades we need to specify this manually to have access to those settings. Not sure that we need it

SteveSugden: What are we using it for?  Apart from importing it I don't see it being used.

PatrykRudnicki: For `commonSettings`

```
lazy val commonSettings: Seq[Setting[_]] = scalaSettings ++
   publishingSettings ++
   defaultSettings() ++
  gitStampSettings
```

In older version of some plugin (probably sbt-distributables) git stamp was inside. With newer version they probably removed it, so we need to manually add dependency and import for it.

Not sure if we need git stamp, but upgrade IMO should be consistent with previous version. I'll try to figure it out if it's necessary setting or we can skip it.

PR #72, opened 2020-01-03
CEDS-1903 Use Instant as caring class for date and time in movement request.

bambuchaAdm:
We use here default ISO_INSTANT DateTimeFormatter - ISO 8601 with milisecond resolution.
Any problems with format are now reported as json validation errros.

We change to change comparison between XMLs because it does not recognize correctly
equality between two dates even if they are point same value after formatting.
After this tests were moved to compare result of `toString` on Nodes


PatrykRudnicki:
One spec: `component.uk.gov.hmrc.exports.movements.RetrospectiveArrivalSpec` failed

bambuchaAdm:
No it should not - it just makes more restrictive on inputs. Should be compatible with old input. 
It:test part :/

PR #78, opened 2020-01-14
CEDS-1412 Add test cases for MUCR association and dissociation

janeturner:


test/component/uk/gov/hmrc/exports/movements/AssociateUCRSpec.scala
```diff
@@ -30,36 +30,72 @@ import uk.gov.hmrc.exports.movements.models.submissions.{ActionType, Submission}
...
+      }
+
+      "used for MUCR Association" in {
+        // Given
+        givenIleApiAcceptsTheSubmission("conversation-id")
+
+        // When
+        val response = post(
+          routes.ConsolidationController.submitConsolidation(),
+          Json.obj("providerId" -> "pid", "eori" -> "eori", "consolidationType" -> "ASSOCIATE_MUCR", "mucr" -> "MUCR", "ucr" -> "MUCR")
```
MaciejRewera: I'd suggest to distinguish between two MUCRs. In this scenario you are associating the eMUCR with itself which is awkward.

MaciejRewera: And what is even more important, you are not able to detect an error if you have switched those two in the process.

janeturner: Good point. Could call the one to associate `associateMUCR`?

MaciejRewera: I'd call it just `MUCR_2`. But it's up to you.

PR #79, opened 2020-01-17
CEDS-2106 refactoring and preparation

MaciejRewera:


test/unit/uk/gov/hmrc/exports/movements/base/CustomsExportsBaseSpec.scala
```diff
@@ -76,7 +76,7 @@ trait CustomsExportsBaseSpec
 
   def messagesApi: MessagesApi = injector.instanceOf[MessagesApi]
 
-  def injector: Injector = app.injector
+  def injector: inject.Injector = app.injector
```
PatrykRudnicki: Why this change? Still it should recognise `Injector`

bambuchaAdm: name clash with test util probably

bambuchaAdm: We found this trait is not used anywhere - we would remove it

test/unit/uk/gov/hmrc/exports/movements/base/Injector.scala
```diff
@@ -0,0 +1,38 @@
...
+  /**
+    * Clearing shared metrics registries to avoid `A metric named jvm.attribute.vendor already exists` error.
+    *
+    * It appears very often with places with injector. This is enough solution for this problem.
+    *
+    * Reference and other solutions: https://github.com/kenshoo/metrics-play/issues/74
+    */
+  SharedMetricRegistries.clear()
+
+  private val injector = GuiceApplicationBuilder().injector()
```
bambuchaAdm: Let's move clear declaration to varaible initialization - i have always problem with those that its order is not really known
```
private val injector =  {
  SharedMetricRegistries.clear()
  GuiceApplicationBuilder().injector()
}
```

PR #85, opened 2020-01-27
CEDS-2091 fetch query responses

MaciejRewera:
This PR does NOT contain implementation for query timeout. It is to follow in the next PR.

### DIFF
--- a/app/uk/gov/hmrc/exports/movements/misc/IleQueryTimeoutCalculator.scala
+++ b/app/uk/gov/hmrc/exports/movements/misc/IleQueryTimeoutCalculator.scala
@@ -14,11 +14,11 @@
  * limitations under the License.
  */
 
-package uk.gov.hmrc.exports.movements.services
+package uk.gov.hmrc.exports.movements.misc
 
 import uk.gov.hmrc.exports.movements.models.submissions.Submission
 
 class IleQueryTimeoutCalculator {
 
-  def hasQueryTimedOut(submission: Submission): Boolean = ???
+  def hasQueryTimedOut(submission: Submission): Boolean = false
 }
--- a/app/uk/gov/hmrc/exports/movements/misc/IleQueryTimeoutCalculator.scala
+++ b/app/uk/gov/hmrc/exports/movements/misc/IleQueryTimeoutCalculator.scala
@@ -20,5 +20,6 @@ import uk.gov.hmrc.exports.movements.models.submissions.Submission
 
 class IleQueryTimeoutCalculator {
 
+  // TODO: Implement this logic with the next opportunity
   def hasQueryTimedOut(submission: Submission): Boolean = false
 }
--- a/test/unit/uk/gov/hmrc/exports/movements/controllers/FakeRequestFactory.scala
+++ b/test/unit/uk/gov/hmrc/exports/movements/controllers/FakeRequestFactory.scala
@@ -16,6 +16,8 @@
 
 package unit.uk.gov.hmrc.exports.movements.controllers
 
+import akka.actor.ActorSystem
+import akka.stream.ActorMaterializer
 import play.api.libs.json.Writes
 import play.api.mvc.{AnyContent, AnyContentAsJson, AnyContentAsXml}
 import play.api.test.FakeRequest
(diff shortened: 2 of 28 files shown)


app/uk/gov/hmrc/exports/movements/misc/IleQueryTimeoutCalculator.scala
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.exports.movements.misc
+
+import uk.gov.hmrc.exports.movements.models.submissions.Submission
+
+class IleQueryTimeoutCalculator {
+
+  def hasQueryTimedOut(submission: Submission): Boolean = false
```
PatrykRudnicki: Can you add TODO that this functionality will be added in the nearest future?

test/unit/uk/gov/hmrc/exports/movements/controllers/FakeRequestFactory.scala
```diff
@@ -26,11 +28,16 @@ import scala.xml.Elem
 
 object FakeRequestFactory {
 
+  lazy val actorSystem: ActorSystem = ActorSystem()
+  lazy val materializer: ActorMaterializer = ActorMaterializer()(actorSystem)
```
PatrykRudnicki: Not sure that there is any sense to create this in our `FakeRequestFactory`

test/unit/uk/gov/hmrc/exports/movements/controllers/NotificationControllerSpec.scala
```diff
@@ -49,8 +49,8 @@ class NotificationControllerSpec extends WordSpec with MustMatchers with Mockito
   private val movementsMetrics = buildMovementsMetricsMock
 
   private val controllerComponents: ControllerComponents = instanceOf[ControllerComponents]
-  implicit private val actorSystem: ActorSystem = ActorSystem()
-  implicit private val materializer: ActorMaterializer = ActorMaterializer()
+  implicit private val actorSystem: ActorSystem = FakeRequestFactory.actorSystem
+  implicit private val materializer: ActorMaterializer = FakeRequestFactory.materializer
```
PatrykRudnicki: Specially like the usage of it require creating/invoking it.

test/unit/uk/gov/hmrc/exports/movements/controllers/IleQueryControllerSpec.scala
```diff
@@ -17,28 +17,34 @@
...
 import scala.concurrent.ExecutionContext.global
 import scala.concurrent.Future
 
 class IleQueryControllerSpec extends WordSpec with MustMatchers with MockitoSugar with ScalaFutures with BeforeAndAfterEach {
 
   private val ileQueryService = mock[IleQueryService]
-
   private val controller = new IleQueryController(ileQueryService, stubControllerComponents())(global)
 
+  implicit val defaultPatience: PatienceConfig = PatienceConfig(timeout = Span(5, Seconds), interval = Span(10, Millis))
```
SteveSugden: Is this used anywhere?

test/unit/uk/gov/hmrc/exports/movements/controllers/NotificationControllerSpec.scala
SteveSugden: What are these Actor thingies used for?  Why do the FE controller specs not use them?

test/unit/uk/gov/hmrc/exports/movements/controllers/FakeRequestFactory.scala
MaciejRewera: The story behind this is that in `FakeRequestFactory` there is a method creating `FakeRequest[AnyContentAsJson]`. And this requires `materializer`. So since the need for it comes from the request factory I thought it is natural to put it here. And did not have better idea to be fair.

test/unit/uk/gov/hmrc/exports/movements/controllers/NotificationControllerSpec.scala
MaciejRewera: I did not want to make it implicit by default, but to create single instances of `ActorSystem` and `ActorMaterializer` in the response factory and only re-use them where needed.

test/unit/uk/gov/hmrc/exports/movements/controllers/IleQueryControllerSpec.scala
MaciejRewera: Implicitly it is used in all `futureValue` calls. I've noticed this being used in `SubmissionControllerSpec` and decided to use it here as well. The reason being that one test was not able to get a response from the controller before the default timeout.

PR #93, opened 2020-02-11
CEDS-2111 Add Mongobee as MongoDB db migration tool

PauloPortugal:


PauloPortugal:
> What about rest of the repositories? We have as I remember four repositories there. Is it enough to update movement submission repository?

Good point. I thought about that, but it does not need to add it now. 
the `dbchangelog` collection is not specific to any collection, so any collection change will be added there. 
For another collection we could create a new `NotificationChangeLog` for notifications and add our changesets there. Probably in microservices where we have multiple collections we need to use the `ChangeSet.id` to identifiy to what collection the change refers to

PR #94, opened 2020-02-11
Ceds 1873- Updated Client-id field in Movements Backend config

alsahid1993:


app/uk/gov/hmrc/exports/movements/config/AppConfig.scala
```diff
@@ -38,8 +38,8 @@ class AppConfig @Inject()(runModeConfiguration: Configuration, servicesConfig: S
 
   def clientIdInventory(implicit hc: HeaderCarrier): String = {
     val userAgent = hc.headers.find(_._1.toLowerCase() == "user-agent").map(_._2).getOrElse {
-      logger.warn("Request had missing User-Agent header. Falling Back to a default Client ID")
-      "default"
+      logger.warn("Request had missing User-Agent header.")
+      throw MissingClientIDException("User Agent")
```
SteveSugden: I don't think you technically need to throw that extra exception.  If you returned an empty string the following line will throw a similar exception.  However, this is ok too.

You could probably use "getOrThrow" (instead of getOrElse) and skip the logging of the warning - no point logging of you are going to stop the app starting with the exception.

test/integration/uk/gov/hmrc/exports/movements/services/SubmissionServiceSpec.scala
```diff
@@ -62,15 +62,15 @@ class SubmissionServiceSpec
...
+          "microservice.services.customs-inventory-linking-exports.client-id.customs-movements-frontend" -> CustomsMovementsAPIConfig.clientId
         )
       )
       .build()
   }
 
   private lazy val movementsService = app.injector.instanceOf[SubmissionService]
 
-  private implicit val hc: HeaderCarrier = HeaderCarrier()
+  private implicit val hc: HeaderCarrier = HeaderCarrier(extraHeaders = Seq("user-agent" -> "customs-movements-frontend"))
```
MaciejRewera: 👍

app/uk/gov/hmrc/exports/movements/config/AppConfig.scala
PatrykRudnicki: I help Alsahid with this implementation. If you use empty string, this empty string will be putted in the configuration and application will try to retrieve the incorrect key. We need to throw an exception, because this case is incorrect

MaciejRewera: I think it would be good to log the service name we can't find client-id for.

app/uk/gov/hmrc/exports/movements/config/AppConfig.scala
```diff
@@ -38,8 +38,16 @@ class AppConfig @Inject()(runModeConfiguration: Configuration, servicesConfig: S
...
+    }
+    servicesConfig.getConfString(s"customs-inventory-linking-exports.client-id.$userAgent", throw MissingClientIDException(userAgent))
+  }
+
+  def clientIdInventoryConfig(implicit hc: HeaderCarrier): String = {
```
SteveSugden: What's this new method for?  Looks to be the same as the one above and only called by a test.  Or have I missed something?

PR #98, opened 2020-02-18
CEDS-2170 Change GatewayTimeout to FailedDependency status for timeout

MaciejRewera:
There was a problem related to sending 5xx HTTP responses between our
services. Any 5xx code triggers Pager Duty Alert which is not the
behaviour we would like to have in this scenario.

Initially the idea was to create a new IleQueryResponseExchangeData
class for the purpose of indicating the timeout. However, this is
essentially a payload driven logic and it is better to drive logic by
HTTP codes.

424 Failed Dependency - seems to be the most appropriate in this
situation among all the codes available: "The request failed because it
depended on another request and that request failed"


SteveSugden:
Just looked up the definition of a 424.
424 FAILED DEPENDENCY
The method could not be performed on the resource because the requested action depended on another action and that action failed.

As you said, it seems a reasonable one to use.

PR #99, opened 2020-03-18
BAU - Add mongock as a DB migration tool

PauloPortugal:
https://github.com/cloudyrock/mongock is an updated fork from Mongobee

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -36,6 +36,7 @@ lazy val microservice = Project(appName, file("."))
   .settings(
     libraryDependencies ++= (AppDependencies.compile ++ AppDependencies.test()).map(_.withSources),
     dependencyOverrides ++= AppDependencies.jettyOverrides,
+    dependencyOverrides += "org.mongodb" % "mongo-java-driver" % "3.12.2",
     evictionWarningOptions in update := EvictionWarningOptions.default.withWarnScalaVersionEviction(false),
     majorVersion := 0,
     scalaVersion := "2.12.8"
--- a/project/AppDependencies.scala
+++ b/project/AppDependencies.scala
@@ -8,15 +8,16 @@ object AppDependencies {
   private val testScope = "test,it"
 
   val compile = Seq(
-    "uk.gov.hmrc"           %% "simple-reactivemongo"       % "7.22.0-play-26",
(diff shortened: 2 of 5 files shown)


project/AppDependencies.scala
```diff
@@ -8,15 +8,15 @@ object AppDependencies {
...
-    "com.github.tototoshi"  %% "scala-csv"                  % "1.3.6",
-    "uk.gov.hmrc"           %% "play-json-union-formatter"  % "1.7.0",
-    "com.github.mongobee"   %  "mongobee"                   % "0.13"
+    "uk.gov.hmrc"                     %% "bootstrap-play-26"          % "1.3.0",
+    "uk.gov.hmrc"                     %% "wco-dec"                    % "0.33.0",
+    "uk.gov.hmrc"                     %% "logback-json-logger"        % "4.6.0",
+    "com.typesafe.play"               %% "play-json-joda"             % "2.6.13",
+    "com.github.tototoshi"            %% "scala-csv"                  % "1.3.6",
+    "uk.gov.hmrc"                     %% "play-json-union-formatter"  % "1.7.0",
+    "com.github.cloudyrock.mongock"   %  "mongock-core"               % "2.0.2"
```
PatrykRudnicki: Do you need add:
`org.mongodb" %  "mongo-java-driver" % "3.12.1`

PauloPortugal: Yes and we also need to override.
Although it worked correctly, but you are spot on.

build.sbt
```diff
@@ -36,6 +36,7 @@ lazy val microservice = Project(appName, file("."))
   .settings(
     libraryDependencies ++= (AppDependencies.compile ++ AppDependencies.test()).map(_.withSources),
     dependencyOverrides ++= AppDependencies.jettyOverrides,
+    dependencyOverrides += "org.mongodb" % "mongo-java-driver" % "3.12.1",
```
PatrykRudnicki: Are you sure that overriding is necessary? `customs-declare-exports` doesn't have it. Should I add this to service?

PauloPortugal: It supports all MongoDB versions, so this avoids us having problems in the near and medium term
https://docs.mongodb.com/ecosystem/drivers/java/

PR #101, opened 2020-04-09
BAU Change response message

PatrykRudnicki:


PatrykRudnicki:
I think it's enough to log the Conversation ID, not put this in the response, but not sure about it.

I can add conv ID to the response

PR #103, opened 2020-07-13
CEDS-2607 Libraries upgrade

PatrykRudnicki:


project/AppDependencies.scala
```diff
@@ -8,23 +8,23 @@ object AppDependencies {
...
-  def test(scope: String = "test") = Seq(
-    "org.scalatest"           %% "scalatest"          % "3.0.8"             % scope,
+  val test = Seq(
+    "org.scalatest"           %% "scalatest"          % "3.0.8"             % "test",
     "org.scalatestplus.play"  %% "scalatestplus-play" % "3.1.2"             % "test",
-    "org.pegdown"             %  "pegdown"            % "1.6.0"             % scope,
-    "com.typesafe.play"       %% "play-test"          % PlayVersion.current % scope,
+    "org.pegdown"             %  "pegdown"            % "1.6.0"             % "test",
+    "com.typesafe.play"       %% "play-test"          % PlayVersion.current % "test",
     "com.github.tomakehurst"  %  "wiremock"           % wireMockVersion     % testScope,
```
SteveSugden, line 28: Just a question - why do we need a constant for `testScope`?

PatrykRudnicki: Hmm, not sure. I can inline this there.

In general I see pattern that for some scopes we're using variables.

I can change it here :)

SteveSugden: No, that' fine.  Just wondered if there was a special reason.

PR #106, opened 2020-07-23
CEDS-2581 Add support for DUCR Part in the Submission model

PatrykRudnicki:


app/uk/gov/hmrc/exports/movements/models/submissions/ActionType.scala
```diff
@@ -25,16 +25,25 @@ object ActionType {
   case object IleQuery extends ActionType("IleQuery")
 
   abstract class ConsolidationType(override val typeName: String, val ileCode: String) extends ActionType(typeName)
+
```
PatrykRudnicki, line 28: I used IntelliJ formatter for this file by mistake

app/uk/gov/hmrc/exports/movements/services/UcrBlockBuilder.scala
```diff
@@ -81,11 +81,11 @@ class UcrBlockBuilder {
     case MucrAssociation | MucrDisassociation                                                => "M"
   }
 
-  def extractUcrBlocksFrom(nodeSeq: NodeSeq): Seq[UcrBlock] = {
+  def extractUcrBlocksForSubmissionFrom(nodeSeq: NodeSeq): Seq[UcrBlock] = {
     val ucrBlocks = (nodeSeq \ XmlTags.ucrBlock).map { node =>
       val ucr = (node \ XmlTags.ucr).text
-      val ucrType = (node \ XmlTags.ucrType).text
       val ucrPartNo = StringOption((node \ XmlTags.ucrPartNo).text)
+      val ucrType = if (ucrPartNo.exists(_.nonEmpty)) DucrPart.codeValue else (node \ XmlTags.ucrType).text
```
PatrykRudnicki, line 88: From XML perspective still the values are `D` and `M`, so we need to recognise which requests are a DUCR Parts requests. This is only for Submission mode and previously has been used only for Submission, so I changed the name to include Submission.

If `ucrPartNo` is defined it means that user sent DUCR Parts request.

PatrykRudnicki: I think that this method can live in SubmissionsFactory class, because it's for Submission purpose, but didn't want to make unnecessary refactor.

SteveSugden: So the logic is...
If there is a ucrPartNo then the type is DucrPart otherwise whatever value is on Xml tag ucrType

PatrykRudnicki: Yes, it can be MUCR or DUCR in the Xml tag

PR #107, opened 2020-08-12
CEDS-2669 - return 500 error if submission fails

SteveSugden:


app/uk/gov/hmrc/exports/movements/controllers/MovementsController.scala
```diff
@@ -30,6 +30,8 @@ class MovementsController @Inject()(submissionService: SubmissionService, overri
 ) extends BackendController(controllerComponents) {
 
   def createMovement(): Action[MovementsExchange] = Action.async(parse.json[MovementsExchange]) { implicit request =>
-    submissionService.submit(request.body).map(_ => Accepted(request.body))
+    submissionService.submit(request.body).map(_ => Accepted(request.body)) recover {
```
PatrykRudnicki, line 33: I'm not sure if the approach with recovering throwable is a good approach. I would say to convert Submission service to return different responses with http status instead of throwing an exception and based on that status return different status to frontend.

SteveSugden: I was just basing this solution on what the service currently does - it returns a Future.failed.
I was surprised that the MovementsController had to do anything to be honest.  Currently that Future.failed is just ignored?

PatrykRudnicki: Future.failed is like throwing an exception so it will work in the same way, but in general I think that the backend should just forward the response with status etc and frontend will need to decide what should be displayed to the user based on the response.

Following the current service implementation I think this will be enough, but for this specific case we probably can consider refactoring the submission service

app/uk/gov/hmrc/exports/movements/services/SubmissionService.scala
```diff
@@ -54,8 +54,10 @@ class SubmissionService @Inject()(
 
       case CustomsInventoryLinkingResponse(status, conversationId) =>
         logger.warn(s"Movement Submission failed with conversation-id=[$conversationId] and status [$status]")
-        Future.failed(
-          new CustomsInventoryLinkingUpstreamException(status, conversationId, "Non Accepted status returned by Customs Inventory Linking Exports")
+        throw new CustomsInventoryLinkingUpstreamException(
```
SteveSugden, line 57: This new approach is based on what IleQueryService does.  Throwing the exception, rather than wrapping it in a future, causes a "500" error to be returned from the controller without any further code changes.

MaciejRewera:
I am wondering if some information on why was the Submission not accepted should be propagated? Because as I understand, ATM we will log the status in the BE, and the FE will receive information that "Something went wrong", but there is no way for the FE to know why?
If this is not a concern, then this implementation should be enough. +1 for logging some more info ;-)

PR #110, opened 2020-11-02
CEDS-2767 Adding new 'internal-user-eori' key into app config…

glidester:
…to be used as the EORI value for internal users

### DIFF
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -170,6 +170,7 @@ microservice {
                 customs-exports-internal-frontend = "5c68d3b5-d8a7-4212-8688-6b67f18bbce7"
             }
             schema-file-path = "conf/schemas/exports/inventoryLinkingResponseExternal.xsd"
+            internal-user-eori = "ABC123"
         }
     }
 }
--- a/test/unit/uk/gov/hmrc/exports/movements/config/AppConfigSpec.scala
+++ b/test/unit/uk/gov/hmrc/exports/movements/config/AppConfigSpec.scala
@@ -121,5 +121,11 @@ class AppConfigSpec extends UnitSpec with MockitoSugar {
       val duration = serviceConfig.ileQueryResponseTimeout
       duration shouldBe Duration.of(30, ChronoUnit.SECONDS)
     }
+
+    "contain correct value for the internal user EORI" in {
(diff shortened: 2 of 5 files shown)


conf/application.conf
```diff
@@ -170,6 +170,7 @@ microservice {
                 customs-exports-internal-frontend = "5c68d3b5-d8a7-4212-8688-6b67f18bbce7"
             }
             schema-file-path = "conf/schemas/exports/inventoryLinkingResponseExternal.xsd"
+            internal-user-eori = "ABC123"
```
MaciejRewera, line 173: What do you think about having separate fields for Badge and Submitter Identifiers? And to name them accordingly? I am just not sure if these are always equal to EORI and in "normal" circumstances they contin different values probably.
Might be worth confirming with @bethany-kish

test/unit/uk/gov/hmrc/exports/movements/config/AppConfigSpec.scala
```diff
@@ -97,6 +98,9 @@ class AppConfigSpec extends UnitSpec with MockitoSugar {
 
       val caught3: Exception = intercept[Exception](configService.sendArrivalUrlSuffix)
       caught3.getMessage shouldBe "Missing configuration for Customs Inventory Linking send arrival URI"
+
+      val caught4: Exception = intercept[Exception](configService.internalUserEori)
```
MaciejRewera, line 102: We usually also add a test for fetching the value from config. Can you add it please?

conf/application.conf
bethany-kish: I just tried an external submission and it didn't send a badge identifier to ILE - so no idea what it would be 'normally'

glidester: My view is that atm the current solution is that these two values are always the same. So unless there is a requirement to be able to set them independently we should not implement it as it might create new opportunities for the config to break the service in a particular environment.

glidester: ... but will ask @bethany-kish to confirm

test/unit/uk/gov/hmrc/exports/movements/config/AppConfigSpec.scala
glidester: Done

MaciejRewera (approved):
LGTM

PR #113, opened 2020-12-03
BAU Implement Exports Migration Tool

MaciejRewera:
This commit only puts-in the tool's code, but migrations are still
executed with Mongock. This can be later changed in MigrationRunner.

### DIFF
--- a/app/uk/gov/hmrc/exports/movements/migrations/changelogs/cache/ChangePackageInformationId.scala
+++ b/app/uk/gov/hmrc/exports/movements/migrations/changelogs/cache/ChangePackageInformationId.scala
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2020 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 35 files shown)


glidester (approved):

PauloPortugal:
I still have mixed feelings about this, but I do not have the energy or time to invest here.
https://www.mongock.io/ seems still the best approach, although we have had some limitations here.

app/uk/gov/hmrc/exports/movements/migrations/changelogs/cache/ChangePackageInformationId.scala
```diff
@@ -0,0 +1,106 @@
+/*
```
PauloPortugal, line 1: Do we need to have commented code?

PauloPortugal (approved):
This is a huge additional cost for the project, although the model/DB record changes will not occur that often from now on-wards.
The cost of understanding this code is worth reconsidering the need for this in Movements.
I do understand the need on Exports, as we may have deep nested objects but in Movements I cannot see this limitation, probably there is one.

Also Mongock may grow it's  community base and they may fix some of the issues.

I feel like that guy that changed religion for his partner. Makes them happy, but I am still questioning the change :)

PR #118, opened 2021-01-13
CEDS-2939 Upgrade to play 2.7

janeturner:


project/AppDependencies.scala
```diff
@@ -8,27 +8,28 @@ object AppDependencies {
...
     "org.mongodb.scala"               %%  "mongo-scala-driver"        % "2.9.0"
   )
 
   val test = Seq(
     "org.scalatest"           %% "scalatest"          % "3.0.8"             % "test",
-    "org.scalatestplus.play"  %% "scalatestplus-play" % "3.1.3"             % "test",
+    "org.scalatestplus.play"  %% "scalatestplus-play" % "4.0.3"             % "test",
     "org.pegdown"             %  "pegdown"            % "1.6.0"             % "test",
     "com.typesafe.play"       %% "play-test"          % PlayVersion.current % "test",
     "com.github.tomakehurst"  %  "wiremock"           % wireMockVersion     % testScope,
```
udinobi, line 28: Wiremock version: 2.27.2

project/AppDependencies.scala
```diff
@@ -8,27 +8,28 @@ object AppDependencies {
...
     "com.typesafe.play"               %% "play-json-joda"             % "2.6.13",
     "com.github.tototoshi"            %% "scala-csv"                  % "1.3.6",
-    "uk.gov.hmrc"                     %% "play-json-union-formatter"  % "1.10.0-play-26",
+    "uk.gov.hmrc"                     %% "play-json-union-formatter"  % "1.12.0-play-27",
     "com.github.cloudyrock.mongock"   %  "mongock-core"               % "2.0.2",
     "org.mongodb.scala"               %%  "mongo-scala-driver"        % "2.9.0"
   )
 
   val test = Seq(
     "org.scalatest"           %% "scalatest"          % "3.0.8"             % "test",
```
udinobi, line 24: 3.2.3

janeturner: It doesn't seem possible to use 3.2.3 of scalatest as there are some incompatibilities between this, the version of play and scalatestplus-play, so I've gone back to 3.0.8

udinobi: Indeed. That's because in +3.2.0 a few Scalatest classes were deprecated (e.g. `org.scalatest.WordSpec` and `org.scalatest.MustMatchers`) and replaced by new ones (`org.scalatest.wordspec.AnyWordSpec` and `org.scalatest.matchers.should.Matchers`).

But I understand that this change would be probably challenging.

We can maybe have a quick chat with the other devs to discuss if it's worth to complete the task with this ticket or open a new one, because sooner or later the upgrade will have to be done.

udinobi (approved):

test/component/uk/gov/hmrc/exports/movements/IleQuerySpec.scala
```diff
@@ -44,46 +44,49 @@ class IleQuerySpec extends ComponentSpec {
...
-          goodsItem = Seq(GoodsItemInfo(totalPackages = Some(13))),
-          movements = Seq(
-            MovementInfo(
-              messageCode = "message-code",
-              goodsLocation = "goods-location",
-              movementDateTime = Some(Instant.parse("2019-12-23T11:40:00.000Z")),
-              movementReference = Some("movement-reference"),
-              transportDetails =
-                Some(Transport(modeOfTransport = Some("mode"), nationality = Some("nationality"), transportId = Some("transport-id")))
+    data = Some(
```
glidester, line 50: Why is this field being changed to an Optional value due to the play upgrade? Looks correct for the Notification class but just wondering how the test passed before this change? :thinking:

janeturner: @glidester Don't really understand this one myself actually, the Notification class hasn't changed at all from master, it doesn't take a `responseType` parameter and the data field is an Option, but I don't understand why they would have passed before as this doesn't seem related to my changes

glidester (approved):

PR #126, opened 2022-05-10
CEDS-3827 Migration to hmrc-mongo from simple-reactivemongo

udinobi:


build.sbt
```diff
@@ -17,8 +17,7 @@ lazy val allTest = Seq(testAll := (test in ComponentTest)
 lazy val microservice = Project(appName, file("."))
   .enablePlugins(PlayScala, SbtAutoBuildPlugin, SbtDistributablesPlugin)
   .settings(
-    libraryDependencies ++= AppDependencies.compile ++ AppDependencies.test,
-    dependencyOverrides ++= AppDependencies.jettyOverrides.toSeq,
+    libraryDependencies ++= Dependencies.compile ++ Dependencies.test,
```
glidester, line 20: Why have we renamed `AppDependencies` to `Dependencies`??? We call it `AppDependencies` in our other projects so this seems very odd!

glidester (changes requested):
Looks good apart from the AppDependencies rename. Would be good if you could take us through this migration at the next dev catchup :+1:

build.sbt
udinobi: It seems to me that the `App` prefix is really superfluous. And in addition to a lot of Hmrc services in github already using 'Dependencies.scala', we (exports) also have services using it...

- cds-exports-acceptance/project/Dependencies.scala
- cds-file-upload-acceptance-tests/project/Dependencies.scala
- cds-file-upload-frontend/project/Dependencies.scala
- cds-file-upload-performance-tests/project/Dependencies.scala
- cds-file-upload/project/Dependencies.scala
- customs-declare-exports-movements/project/Dependencies.scala
- exports-declarations-performance-tests/project/Dependencies.scala

udinobi:
> Would be good if you could take us through this migration at the next dev catchup +1

Of course, np.

build.sbt
glidester: So apart from SFUS (that our team inherited) and the test repos all our main services use 'AppDependencies' then. I just don't see the point in making this change and brining in an inconsistency between the Movement frontend repos and the BE. Its not like this change will bring consistency across all our repos. Maybe if you are really keen on changing to this name we can do it as part of the lib upgrade tech debt tickets 3908 - 3911?

glidester (approved):

PR #145, opened 2024-05-14
CEDS-5803 Increase the parsable size of received Notification payloads

udinobi:


glidester (changes requested):
Looks fine part from the fact that `maxNotificationPayloadSize` is an optional value and not having a default value specified in application.conf. 
I think having this as a non-optional with a default set in application.conf will be easier for others to understand (and will avoid the config warning that `max.notification.payload.size` overrides nothing when we add it in production :wink:

